PR #7, opened 2016-01-05
Fixed file names under dist/* to match bower.json

xaka:


BigstickCarpet:
Thanks for pointing this out.  The file name was actually wrong in the `bower.json` file.  It should have been `ref-parser.js`, as mentioned [in the ReadMe file](https://github.com/BigstickCarpet/json-schema-ref-parser#web-browsers)

xaka:
I guess either way is fine. It's just usually files are named after package. Anyway, thanks!

BigstickCarpet:
yeah, it was named after the package originally, but I decided that the was too long (`bower_components/json-schema-ref-parser/json-schema-ref-parser.js`), so I shortened it.  :)

PR #8, opened 2016-01-05
Allow users to override/replace/add resolvers and parsers

xaka:


BigstickCarpet:
Great minds think alike!  I'm literally working on implementing this **right now**.  Instead of just two options (one for files and one for urls), you'll have the ability to define any number of arbitrary resolvers.  This will allow people to support any protocol they need, such as `ftp://`, `mongodb://`, etc.

Hopefully I'll finish this tonight.

xaka:
That's great! Please keep door open for substituting our own generic resolver regardless of protocol. In my case I have a virtual filesystem within browser and I wanted to be able to resolve referenced documents from there.

BigstickCarpet:
:+1: That should be totally possible.  I'm allowing you to completely override and/or disable any of the built-in resolvers, so you can completely disable the default functionality and substitute your own if you want.

BigstickCarpet:
Once I get the code committed, I'd love your feedback.  I want to make sure it's flexible enough for a variety of different use-cases.

xaka:
Sure thing! You can open PR when you're done and I'd take a look at proposed changes with pleasure. Let's close this one.

xaka:
I've just got one important thought. With overridable resolvers there has to be an option to disable caching entirely otherwise resolvers are going to work with unexpected side effects (and usually have their own caching strategies).

BigstickCarpet:
Yeah, each resolver will have its own cache duration, just like the `options.cache.xxx` that exists today.  There will be default cache durations, but they can be changed by the user.  And any cache duration can be set to zero to disable caching altogether for that resolver.   Does that meet your needs?

xaka:
:+1:

BigstickCarpet:
@xaka - I've check-in my changes (didn't bother doing a new PR).  It's not fully tested yet, so I haven't released it to NPM or anything, and I haven't updated the docs yet.  But feel free to take a look and give your feedback on what's there.

I'm calling it a night.  Will pick it back up tomorrow.

xaka:
I took a look and I think the change is reasonable, and gives enough flexibility to override the default behavior. I'm looking forward for next release in order to jump off my fork as I really need that "overridable resolvers" feature.

malko:
:up:

Nautigsam:
:up: I would love to see this implemented :)

BigstickCarpet:
Done!  This just landed in [v3.0.0](https://github.com/BigstickCarpet/json-schema-ref-parser/blob/master/CHANGELOG.md#v300-2016-04-03)

PR #18, opened 2016-05-12
Using SCHEMAS env var to address issue with local refs

iaincollins:
I ran into a problem with local relative file references and addressed it it with the follow.

I assume you would not want to accept this pull request due to the nature of the 'fix', as it uses a env var as a work around, but I thought I'd make you aware of this issue as someone with familiarity with the code base probably have a better idea of how to address it more elegantly and this explains the issue I ran into.

The problem I had was with local relative file references that are nested.

e.g.

* Define a **Place** schema.
* Define a **Person** schema references **PostalAddress** object in **Place** schema (with $ref: "schemas/Place#/definitions/PostalAddress").
* Define a **Organization** schema references **Person** schema (with $ref: "schemas/Person").

In this case, Person will instatiate correctly, but Organization fails when it tries to load Person as the path to the schema file for the Place object is wrong.

Instead of 'schemas/Person.json#/definitions/PostalAddress' it tries to load it as  'schemas /schemas/Person#/definitions/PostalAddress' .

I am inelegantly solving it by allowing the setting of a SCHEMAS environment variable, which is used as the root path for all reference to local schema files if it is set.

It is possible I've just been doing it wrong (if so I'd appreciate feedback on that) and if not it could still be solved much more nicely if anyone has the time to think through the options. In my case I'm allow people to configure where schemas are located in my app using a SCHEMAS env var in any case, so it just made sense to leverage it.

I appreciate it's also possible to solve it by using *absolute* paths to local files, but think that's undesirable as code would need to be deployed to the same location on every platform.

### DIFF
--- a/lib/dereference.js
+++ b/lib/dereference.js
@@ -96,6 +96,9 @@ function crawl(obj, path, pathFromRoot, parents, $refs, options) {
 function dereference$Ref($ref, path, pathFromRoot, parents, $refs, options) {
   debug('Dereferencing $ref pointer "%s" at %s', $ref.$ref, path);
 
+  if (process.env.SCHEMAS && !/^https?:\/\//i.test($ref.$ref) && !/^\#/.test($ref.$ref))
+    path = process.env.SCHEMAS;
+    
   var $refPath = url.resolve(path, $ref.$ref);
   var pointer = $refs._resolve($refPath, options);
 
--- a/lib/resolve-external.js
+++ b/lib/resolve-external.js
@@ -92,6 +92,9 @@ function crawl(obj, path, $refs, options) {
 function resolve$Ref($ref, path, $refs, options) {
   debug('Resolving $ref pointer "%s" at %s', $ref.$ref, path);
 
+  if (process.env.SCHEMAS && !/^https?:\/\//i.test($ref.$ref) && !/^\#/.test($ref.$ref))
+    path = process.env.SCHEMAS;
+    
   var resolvedPath = url.resolve(path, $ref.$ref);
   var withoutHash = url.stripHash(resolvedPath);
 


lib/dereference.js
```diff
@@ -96,6 +96,9 @@ function crawl(obj, path, pathFromRoot, parents, $refs, options) {
 function dereference$Ref($ref, path, pathFromRoot, parents, $refs, options) {
   debug('Dereferencing $ref pointer "%s" at %s', $ref.$ref, path);
 
+  if (process.env.SCHEMAS && !/^https?:\/\//i.test($ref.$ref) && !/^\#/.test($ref.$ref))
+    path = process.env.SCHEMAS;
```
BigstickCarpet: Codacy Issue found: [Expected '{' and instead saw 'path'.](https://www.codacy.com/app/jamesmessinger/json-schema-ref-parser/file/2557103956/issues/source?bid=2491852&fileBranchId=3292219#l100)

lib/resolve-external.js
```diff
@@ -92,6 +92,9 @@ function crawl(obj, path, $refs, options) {
 function resolve$Ref($ref, path, $refs, options) {
   debug('Resolving $ref pointer "%s" at %s', $ref.$ref, path);
 
+  if (process.env.SCHEMAS && !/^https?:\/\//i.test($ref.$ref) && !/^\#/.test($ref.$ref))
+    path = process.env.SCHEMAS;
```
BigstickCarpet: Codacy Issue found: [Expected '{' and instead saw 'path'.](https://www.codacy.com/app/jamesmessinger/json-schema-ref-parser/file/2557103967/issues/source?bid=2491852&fileBranchId=3292219#l96)

BigstickCarpet:
Thanks for the PR, but I want to make sure I understand the problem first.  Can you post some sample files that reproduce the problem?  Based on your explanation above, it sounds like the paths in your `$ref` pointers might be wrong, but I'm not sure.   

Remember that all JSON Reference paths are [relative the referring file](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03#section-4).  So if `file-a.json` and `file-b.json` are in the same directory, then any of the following JSON References would work:

**file-a.json**:
```json
{
  "relative reference": {
    "$ref": "file-b.json"
  },
  "relative reference (from current directory)": {
    "$ref": "./file-b.json"
  },
  "relative reference (up a directory, and then back down)": {
    "$ref": "../parent-dir/file-b.json"
  },
  "absolute reference": {
    "$ref": "/absolute/path/to/parent-dir/file-b.json"
  }
}
```

iaincollins:
Hi there, thanks for the response - and for a great library, which works amazingly well.

I tried using the both just the name and ./ in front of it in the file in the $ref, but in the scenario described above it failed to deference the 'Organization' object as the path seemed to be determined incorrectly when it tried to deferences in the Place ref in the Person schema.

(But just dereferencing one level down, with a Person object worked fine.)

It *seems* like it's actually bug in now the base path is determined but wasn't 100% sure and couldn't figure out how to address it (I tried but it was messy and so fell back to this).

For reference, the schemas I am using are here, they are all in the same directory:

https://github.com/glitchdigital/structured-data-api/blob/master/schemas/Place.json
https://github.com/glitchdigital/structured-data-api/blob/master/schemas/Person.json
https://github.com/glitchdigital/structured-data-api/blob/master/schemas/Organization.json

The code which uses json-schema-ref-parser is here:

https://github.com/glitchdigital/structured-data-api/blob/master/lib/schemas.js

Some hairy stuff happens after it loads the schema, but before that all it's doing is looping through files in the schema directory. It occurs to me now I am reusing the json-schema-ref-parser object and I'm not sure if any internal caching could trigger it, but I didn't think to test that.

Hopefully I'm not wasting your time… :)

BigstickCarpet:
Ah, ok, I see the problem.  You're calling `refParser.dereference()` with an _object_, not a file path.  If you call it with a file path, then it will automatically resolve everything based on that path.  If you call it with an object, then by default it resolves everything based on the current working directory, unless you specify a file path along with the object.

In your code, you're reading the object directly from a `.json` file and then passing it to `refParser.dereference()`.  There's no need to do that.  Just pass the file path to `refParser.dereference()` and it'll read it (and all the referenced files) for you.

Another option would be to continue reading the `.json` file yourself, and passing _both_ the parsed object _and_ the file path to `refParser.dereference()`, so it can correctly resolve the references in the object.

iaincollins:
Ah! Doh. The code evolved to use the parser over time and it didn't occur to me I was passing an object (and that of course it wouldn't know the path where it the schema on disk if I was passing an object).

I'll close this PR, fix up the code and nuke my fork.

Thanks for your help, and sorry I was was indeed wasting your time :)

BigstickCarpet:
No worries.  I'm glad it's working well for you!

Issue #27, opened 2016-09-06
Module not found: Error: Cannot resolve module 'fs'


rowino:
I am using the babel [generator boilerplate](https://github.com/babel/generator-babel-boilerplate) for a library I am building. I added this library and imported it into my project. Now when I run gulp build I get the following error. Any assistance would be highly appreciated.

```
stream.js:74
      throw er; // Unhandled stream error in pipe.
      ^
Error: ./~/json-schema-ref-parser/lib/resolvers/file.js
Module not found: Error: Cannot resolve module 'fs' in /home/raphael/work/projects/packages/json-schema-form-generator/node_modules/json-schema-ref-parser/lib/resolvers
resolve module fs in /home/raphael/work/projects/packages/json-schema-form-generator/node_modules/json-schema-ref-parser/lib/resolvers
  looking for modules in /home/raphael/work/projects/packages/json-schema-form-generator/node_modules
    /home/raphael/work/projects/packages/json-schema-form-generator/node_modules/fs doesn't exist (module as directory)
    resolve 'file' fs in /home/raphael/work/projects/packages/json-schema-form-generator/node_modules
      resolve file
        /home/raphael/work/projects/packages/json-schema-form-generator/node_modules/fs doesn't exist
        /home/raphael/work/projects/packages/json-schema-form-generator/node_modules/fs.webpack.js doesn't exist
        /home/raphael/work/projects/packages/json-schema-form-generator/node_modules/fs.web.js doesn't exist
        /home/raphael/work/projects/packages/json-schema-form-generator/node_modules/fs.js doesn't exist
        /home/raphael/work/projects/packages/json-schema-form-generator/node_modules/fs.json doesn't exist
[/home/raphael/work/projects/packages/json-schema-form-generator/node_modules/fs]
[/home/raphael/work/projects/packages/json-schema-form-generator/node_modules/fs]
[/home/raphael/work/projects/packages/json-schema-form-generator/node_modules/fs.webpack.js]
[/home/raphael/work/projects/packages/json-schema-form-generator/node_modules/fs.web.js]
[/home/raphael/work/projects/packages/json-schema-form-generator/node_modules/fs.js]
[/home/raphael/work/projects/packages/json-schema-form-generator/node_modules/fs.json]
 @ ./~/json-schema-ref-parser/lib/resolvers/file.js 2:14-27
```

BigstickCarpet:
[...]

PR #30, opened 2016-11-29
adding support for "browser" field, making inclusion in the browser more robust.

sramam:
Adding the "browser" field to package.json.
For details, see: https://github.com/defunctzombie/package-browser-field-spec

Useful when one is not using bower to install this package. Atleast webpack seems to pick this up without issue.
Webpack seems to support it - seems like an elegant solution for mixed usage libraries. 

Seems to be a better solution than #27 https://github.com/BigstickCarpet/json-schema-ref-parser/issues/27

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,9 @@
     "release": "npm run upgrade && npm test && npm run bump && npm publish",
     "start": "npm run watch & http-server -o -c-1"
   },
+  "browser": {
+    "fs": false
+  },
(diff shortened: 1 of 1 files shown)


BigstickCarpet:
@sramam - Thanks for the PR!  much appreciated

BigstickCarpet:
I just released v3.2.0, which includes this PR

PR #34, opened 2017-02-02
Feature/ignore node

jcdarwin:
This pull request provides a new option, `dereference.ignoreNodes`, which specifies an array of node names that should not be dereferenced.

For example, in the case of anyOf nodes, we may want to dereference the references later, as this will allow us to use the reference name, such as when we write a resolve function for GraphQL and we then use the reference name for the type name.

In our use case, the part of the JSON schema would look as follows:

```
        "RelatedAssets": {
            "anyOf": [
                { "$ref": "#/definitions/RelatedGallery" },
                { "$ref": "#/definitions/RelatedHtmlAsset" },
                { "$ref": "#/definitions/RelatedImage" },
                { "$ref": "#/definitions/RelatedListAsset" },
                { "$ref": "#/definitions/RelatedQueryAsset" },
                { "$ref": "#/definitions/RelatedVideo" },
                { "$ref": "#/definitions/RelatedArticle" }
            ]
        },
```

If we dereference the references now, we lose the reference names.

The node names specified by the  `dereference.ignoreNodes` array will not be dereferenced, nor will any children.


pateketrueke:
Hey, I see this also useful for [json-schema-faker](https://github.com/json-schema-faker/json-schema-faker), but could be worthy fixing #40 first.

I'm glad to help, just commenting for keeping this active.

See you later! 🍻

BigstickCarpet:
Hi @jcdarwin,
It seems like the only reason you're wanting to keep the `$ref`s in there is so you can access the name of the type definition (e.g. RelatedGallery, RelatedImage, RelatedVideo, etc.).  But couldn't you get that same information from the `title` attribute on your type definitions?  The `title` attribute is still available after dereferencing, so you can use it to distinguish between different type definitions.

For example:
```json
{
  "RelatedAssets": {
    "anyOf": [
      { "$ref": "#/definitions/RelatedGallery" },
      { "$ref": "#/definitions/RelatedImage" },
      { "$ref": "#/definitions/RelatedVideo" },
    ]
  },
  "definitions": {
    "RelatedGallery": {
      "title": "RelatedGallery",
      "properties": {
        ...
      }
    },
    "RelatedImage": {
      "title": "RelatedImage",
      "properties": {
        ...
      }
    },
    "RelatedVideo": {
      "title": "RelatedVideo",
      "properties": {
        ...
      }
    },
  }
}
```

pateketrueke:
Definitely I go with @BigstickCarpet answer, I proved my self that adding custom properties, or even relying on known/valid ones is fine.

You could try to extract definitions into separated files, and then add a "id" property which actually will do the same.

PR #38, opened 2017-04-24
add typescript typings

bcherny:
It doesn't seem like you use TS (at least in your open source work), but it's TS best practice to commit typings directly to the repo they type.

If you would rather not do that, I can do the next best thing and commit these typings to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped), the central typings repo for TypeScript.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1,0 +1,306 @@
+import { JSONSchema4, JSONSchema4Type } from 'json-schema'
+
+export = $RefParser
+
+/**
+ * This is the default export of JSON Schema $Ref Parser. You can creates instances of this class using new $RefParser(), or you can just call its static methods.
+ *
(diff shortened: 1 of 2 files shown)


BigstickCarpet:
Wow, thanks for the PR!  I don't use TypeScript, so I don't have any easy way to verify that these type definitions are correct, so I guess I'll just assume they are.  :)

bcherny:
@BigstickCarpet I've been using them to a limited extent in https://github.com/bcherny/json-schema-to-typescript. They're correct as far as I tested, and if there are errors (there's always something 😃 ) fixing them is as easy as updating them in this repo (typings have no caching layer besides NPM's, and are versioned along with the package they type).

qm3ster:
@bcherny my fren, it's #100 **all** TypeScript now xD
:100: :100: :100: :100:

PR #47, opened 2017-08-28
added option to allow unresolved refs to be retained in the schema an…

martingarn:
We wanted the ability to report on all of the unresolved refs.  Added an options that would allow that and not exit on the first unresolved ref.


lib/options.js
```diff
@@ -67,7 +67,14 @@ $RefParserOptions.defaults = {
...
      */
-    circular: true
+    circular: true,
+    /**
+     * The default behavior is to throw an error if a reference is not found.
+     * If false, then leave the $ref: value object in the json schema
+     *
+     * @type {boolean}
+     */
+    onErrorThrow: true
```
Piterden: Thanx for comments, friend))

philsturgeon:
To get a handle on this repo we're having to clear house a little. If you think this is still a useful and relevant change please can you rebase and update it? Might be a tricky job, might be easier to start over, it depends how much has changed.

Thank you for your time making this in the first place!

PR #51, opened 2017-10-10
Fix ts typings for resolve.http and resolve.file

RomanGotsiy:
Current ts definitions specify `canRead` as a required property in options. Fixed it using [`Partial`](https://www.typescriptlang.org/docs/handbook/advanced-types.html#mapped-types) mapped type.

@bcherny, could you plz review


BigstickCarpet:
@RomanGotsiy - Thanks for the PR.  I'm not very familiar with TypeScript, so I have no idea if this is the correct type definition.  Any feedback from @bcherny would be much appreciated.  If you guys say it looks good, then I'll merge the PR.

PR #55, opened 2017-10-19
Move `eslint-config-modular` to devDependencies.

brendo:
Just a small tweak 👍


BigstickCarpet:
Good catch!  Thanks for the PR!

PR #62, opened 2017-12-06
fix bundle producing schema with corrupted refs in some cases

RomanGotsiy:
Hey @BigstickCarpet,

ReDoc users found an interesting issue in `json-schema-ref-parser`. This is hard to explain so I will just show an example. If we have two files:

`index.yaml`
```yaml
paths:
  $ref: '#/components'
external:
  go:
    deeper:
      $ref: other.yaml#/thing'
components:
  test:
    $ref: 'other.yaml#/thing'
```

and `other.yaml`:
```yaml
thing:
  type: string
```

bundle produced the following output:

```json
{
  "paths": {
    "$ref": "#/components"
  },
  "external": {
    "go": {
      "deeper": {
        "$ref": "#/paths/test" // <--- broken ref here
      }
    }
  },
  "components": {
    "test": {
      "type": "string"
    }
  }
}
```

It actually depends on the deep level of the ref, I believe it has something to sorting order.

I've fixed this in this PR. The fix seems reasonable to me but I'm not sure if this is correct so please check it carefully. Also added test case but not sure if the names are so good.

### DIFF
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -20,7 +20,7 @@ function bundle (parser, options) {
 
   // Build an inventory of all $ref pointers in the JSON Schema
   var inventory = [];
-  crawl(parser, 'schema', parser.$refs._root$Ref.path + '#', '#', inventory, parser.$refs, options);
+  crawl(parser, 'schema', parser.$refs._root$Ref.path + '#', '#', inventory, parser.$refs, 0, options);
 
   // Remap all $ref pointers
   remap(inventory);
@@ -37,12 +37,12 @@ function bundle (parser, options) {
  * @param {$Refs} $refs
  * @param {$RefParserOptions} options
  */
-function crawl (parent, key, path, pathFromRoot, inventory, $refs, options) {
+function crawl (parent, key, path, pathFromRoot, inventory, $refs, seq, options) {
   var obj = key === null ? parent : parent[key];
 
   if (obj && typeof obj === 'object') {
     if ($Ref.isAllowed$Ref(obj)) {
-      inventory$Ref(parent, key, path, pathFromRoot, inventory, $refs, options);
+      inventory$Ref(parent, key, path, pathFromRoot, inventory, $refs, seq++, options);
     }
     else {
       var keys = Object.keys(obj);
@@ -60,10 +60,10 @@ function crawl (parent, key, path, pathFromRoot, inventory, $refs, options) {
         var value = obj[key];
 
         if ($Ref.isAllowed$Ref(value)) {
-          inventory$Ref(obj, key, path, keyPathFromRoot, inventory, $refs, options);
+          inventory$Ref(obj, key, path, keyPathFromRoot, inventory, $refs, seq++, options);
         }
         else {
-          crawl(obj, key, keyPath, keyPathFromRoot, inventory, $refs, options);
+          crawl(obj, key, keyPath, keyPathFromRoot, inventory, $refs, seq, options);
         }
       });
     }
@@ -82,7 +82,7 @@ function crawl (parent, key, path, pathFromRoot, inventory, $refs, options) {
  * @param {$Refs} $refs
  * @param {$RefParserOptions} options
  */
-function inventory$Ref ($refParent, $refKey, path, pathFromRoot, inventory, $refs, options) {
+function inventory$Ref ($refParent, $refKey, path, pathFromRoot, inventory, $refs, seq, options) {
   if (inventory.some(function (i) { return i.parent === $refParent && i.key === $refKey; })) {
     // This $Ref has already been inventoried, so we don't need to process it again
     return;
@@ -108,11 +108,14 @@ function inventory$Ref ($refParent, $refKey, path, pathFromRoot, inventory, $ref
     value: pointer.value,         // The resolved value of the $ref pointer
     circular: pointer.circular,   // Is this $ref pointer DIRECTLY circular? (i.e. it references itself)
     extended: extended,           // Does this $ref extend its resolved value? (i.e. it has extra properties, in addition to "$ref")
-    external: external            // Does this $ref pointer point to a file other than the main JSON Schema file?
+    external: external,           // Does this $ref pointer point to a file other than the main JSON Schema file?
+    seq: seq                      // Sequence number of this $ref
   });
 
+  var refPathFromRoot = external ? pathFromRoot : $ref.$ref;
+
   // Recursively crawl the resolved value
-  crawl(pointer.value, null, pointer.path, pathFromRoot, inventory, $refs, options);
+  crawl(pointer.value, null, pointer.path, refPathFromRoot, inventory, $refs, seq, options);
 }
 
 /**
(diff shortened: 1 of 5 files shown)


RomanGotsiy:
Sorry, just noticed that the fix doesn't solve all the cases.

I've investigated a bit deeper and it seems something is wrong with sorting here: https://github.com/BigstickCarpet/json-schema-ref-parser/blob/master/lib/bundle.js#L161
I will investigate further, will update this PR and then ping you.

RomanGotsiy:
Seems that sorting is not stable. Check out the following two screen-shots of `console.table`-ing 
`inventory` array at this point: https://github.com/BigstickCarpet/json-schema-ref-parser/blob/master/lib/bundle.js#L166

image

here is same spec, just with one more reference added:

image

Check the order of item 3 and 4 at the first screenshot and 4 and 5 at the second: while all the details are the same except `pathFromRoot` they are in different order (`401` first, `403` next and vice versa).
Because the `pathFromRoot` for all the refs is calculated in order of visiting corresponding items in `inventory` array should be ordered in the same order to not produce wrong pointers.
Updated my PR.

I wasn't able to minimize this use case to small reproducible sample (as it depends somehow on refs number in the array and is just crazy). But here is a repo with the spec that reproduces issue: https://github.com/mikunn/openapi-test-refs

Sorry, maybe I am not clear enough, but this issue is hard to explain for me :(

BigstickCarpet:
Thanks for finding and debugging this, @RomanGotsiy.  I'll take a look at it as soon as I get a chance (probably next week).  As long as it doesn't break existing tests/functionality, I'll merge it.

BigstickCarpet:
Sorry, @RomanGotsiy  - This week has been crazy.  But I'm off next week for the holidays, so I'll have time to take a look at this.

VatroslavByte:
From what I figured, and what works in my case, a simple fix is needed. 
In bundle.js line 175
```
else if (i.file === file && i.hash === hash) {
      // This $ref points to the same value as the prevous $ref, so remap it to the same path
      i.$ref.$ref = pathFromRoot;
}
```
`pathFromRoot` should be replaced with `hash`

```
else if (i.file === file && i.hash === hash) {
      // This $ref points to the same value as the prevous $ref, so remap it to the same path
      i.$ref.$ref = hash;
}
```
(Sorry, I'm not very familiar with pull requests)From what I figured, and what works in my case, a simple fix is needed. 
In bundle.js line 175
```
else if (i.file === file && i.hash === hash) {
      // This $ref points to the same value as the prevous $ref, so remap it to the same path
      i.$ref.$ref = pathFromRoot;
}
```
`pathFromRoot` should be replaced with `hash`

```
else if (i.file === file && i.hash === hash) {
      // This $ref points to the same value as the prevous $ref, so remap it to the same path
      i.$ref.$ref = hash;
}
```
(Sorry, I'm not very familiar with pull requests)

BigstickCarpet:
@RomanGotsiy and @VatroslavByte - 👉 **I need your help to test this, please** 👈

I've [modified the `bundle()` logic](https://github.com/BigstickCarpet/json-schema-ref-parser/commit/352f3f24573625a7a4076ef27f6ecedbd44fc2b1) to account for the number of indirections that are required to resolve each `$ref`.  This change should cause the `bundle()` method to use the most direct reference whenever possible, which _should_ fix the issue you were seeing.

I've published a **beta** version of `json-schema-ref-parser` to npm that contains these changes.  You'll need to install version `4.1.0-beta.1`, which you can do by running the following command:

```
npm install json-schema-ref-parser@beta
```

Please give this a try and let me know if it works for your use-cases.  I really appreciate your help and feedback.

VatroslavByte:
The **beta** didn't fix the issue I was having.

The problem is that I was referencing objects through **root** json file, rather than directly referencing the json file where the object actually is. 
And I can see why it is a problem once you introduce circular dependencies.

The behavior that I was expecting is that it would merge root json with all objects from external files, and refs in external files would be replaced with shortest path to that object in root file.

Example: 
In file `"models/Event.json"` I have a ref to object `"../root.json#/components/schemas/SportId"` (SportId object is also in a separate file), and what I'm expecting is that it is replaced with `"#/components/schemas/SportId"` once` "Event.json"` is merged into `"root.json"`.


The funny thing is, I created an test app that is literally copy/paste of my existing app (same code, jsons and json-schema-ref-parser version), and I get the output which I described in the example above. 
How is it possible to have different outputs with same jsons?


**NOTE:**
When I reference the objects directly where the files are (not through root file), everything works fine.

BigstickCarpet:
@VatroslavByte - thanks for your feedback.  I tried reproducing the issue that you mentioned, but was unable to.  [Here is my reproduction](https://runkit.com/bigstickcarpet/json-schema-ref-parser-issue-62).  If you look at the output, you'll see that the reference to `root.json#/components/schemas/SportId` is replaced with `#/components/schemas/SportId` as expected.  I get the same result with both the old code (v4.0.4) and the new code (v4.1.0-beta.1).

Perhaps I'm missing something and failing to accurately reproduce your example.  If you could post reproduction code, similar to mine above, I'd really appreciate it.

VatroslavByte:
Well, I'm also having trouble reproducing it. That's why I mentioned that it works in one app and not the other. 

I'll try and figure out why that is and whip up an example for you these days.

BigstickCarpet:
I decided to go ahead and release this change publicly as [version 4.1.0](https://github.com/BigstickCarpet/json-schema-ref-parser/blob/master/CHANGELOG.md#v410-2018-01-17), since it definitely fixes the original problem that was pointed out by @RomanGotsiy, and all of my tests show that it improves the algorithm without breaking any existing functionality.

RomanGotsiy:
Hey @BigstickCarpet,

Sorry for the long reply. This seems awesome and is fixing my original problem. Thanks so much!

Also, thanks for the explanation that the produced references are valid. Although I can't find anything about this behavior in specs (would be really great if you can point me to), I checked a few JSON schema implementations and they actually work fine with transitive refs like those.

brettstack:
@BigstickCarpet any plans to fix this for all scenarios? Greatly appreciated.

BigstickCarpet:
@brettstack - Not sure what you mean.  Is there a specific scenario where it doesn't work?  If so, can you post a repro?  AFAIK, this fix solves the problem for all scenarios that I've tested.

brettstack:
Simples repro I could make:

```yaml
definitions:
  Resource:
    properties:
      resourceId:
        $ref: '#/definitions/indirectId'
  indirectId:
    $ref: '#/definitions/id'
  id:
    title: ID
```

yields

```json
{
    "definitions": {
        "Resource": {
            "properties": {
                "resourceId": {
                    "$ref": "#/definitions/id"
                }
            }
        },
        "indirectId": {
            "$ref": "#/definitions/id"
        },
        "id": {
            "title": "ID"
        }
    }
}
```

Expected `definitions.Resource.properties.resourceId.$ref` to be '#/definitions/indirectId'

BigstickCarpet:
@brettstack - In your example above, that's exactly the output that I would expect, especially after applying this fix.  `#/definitions/indirectId` is an _indirect_ reference, whereas `#/definitions/id` is a _direct_ reference, so the `bundle()` method will choose `#/definitions/id`, since it's the most direct reference.

brettstack:
Thanks. What if I wanted to keep the indirect references? Is that possible?

BigstickCarpet:
@brettstack - If you want to leave the references as-is, rather than re-mapping them, then just call [the `.resolve()` method](https://github.com/BigstickCarpet/json-schema-ref-parser/blob/master/docs/ref-parser.md#resolveschema-options-callback) instead of `.bundle()`.   The `.bundle()` method is specifically for re-mapping references.

PR #71, opened 2018-02-05
Improve typings by allowing JSONSchema6 objects

krenor:
I've found no restriction which only applies to `JSONSchema4` objects.

Also `@types/node` have been added to dev-dependencies since `Buffer` can't be found without them.


krenor:
@BigstickCarpet This seems like an Issue with Travis setup regarding yarn.. should I remove the `yarn.lock` file?

BigstickCarpet:
This PR has been released in v4.1.1 🙌

PR #75, opened 2018-03-04
When the parser is unable to resolve a path, give the user supplied path in the error.

realityking:
This is an attempt to fix BigstickCarpet/swagger-parser#76.

I'm a little out of my depth here but as far as I can tell this change should work in all-cases.

### DIFF
--- a/lib/ref.js
+++ b/lib/ref.js
@@ -75,10 +75,11 @@ $Ref.prototype.get = function (path, options) {
  *
  * @param {string} path - The full path being resolved, optionally with a JSON pointer in the hash
  * @param {$RefParserOptions} options
+ * @param {string} [originalPath] - The original, unmodified path being resolved
  * @returns {Pointer}
  */
-$Ref.prototype.resolve = function (path, options) {
-  var pointer = new Pointer(this, path);
+$Ref.prototype.resolve = function (path, options, originalPath) {
+  var pointer = new Pointer(this, path, originalPath);
   return pointer.resolve(this.value, options);
 };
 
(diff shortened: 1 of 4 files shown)


lib/ref.js
```diff
@@ -75,10 +75,11 @@ $Ref.prototype.get = function (path, options) {
  *
  * @param {string} path - The full path being resolved, optionally with a JSON pointer in the hash
  * @param {$RefParserOptions} options
+ * @param {string} [originalPath] - The original, unmodified path being resolved
  * @returns {Pointer}
  */
-$Ref.prototype.resolve = function (path, options) {
-  var pointer = new Pointer(this, path);
+$Ref.prototype.resolve = function (path, options, originalPath) {
```
BigstickCarpet: I don't see anywhere where this method is called with the new parameter.  Am I missing something?  Or is the parameter always `undefined`?  If the latter, then we can remove the parameter from here and from the `Pointer` constructor.

realityking: it's used here https://github.com/BigstickCarpet/json-schema-ref-parser/pull/75/files#diff-37de81b33471a7915002750868466e52R157

BigstickCarpet: thanks.  Not sure how I missed that

BigstickCarpet:
json-schema-ref-parser v5.0.0 has been published to npm and includes this fix.  Thanks again for your help!

alasdairhurst:
Does this also fix the issue with the custom resolver not getting the original $ref?
https://github.com/BigstickCarpet/swagger-parser/issues/76#issuecomment-339014108

PR #77, opened 2018-03-22
set appropriate schema reference if schema id is provided by an object

Azerothian:
At the moment in-browser projects when using parsing inline object schemas, the base URI is being taken from the current URL versus what is actually defined in the schema via the $id tag.

http://json-schema.org/latest/json-schema-core.html#rfc.section.9.2


Azerothian:
still needs some work

PR #79, opened 2018-04-09
chore: moving karma-edge-launcher to dev dependencies.

neverendingqs:
I suspect we don't need `karma-edge-launcher` in production.


BigstickCarpet:
Good catch!  thanks for the PR

PR #80, opened 2018-04-11
Add the friendlyPath argument to $Refs::_resolve

realityking:
Unfortunately #75 wasn't enough to get everything back to how it worked in 3.x 😢 With this change the tests for apiaryio/fury-adapter-swagger#158 finally pass.

I don't know this library well enough to write a test for this change. Any pointer would be appreciated.

### DIFF
--- a/lib/dereference.js
+++ b/lib/dereference.js
@@ -97,7 +97,7 @@ function dereference$Ref ($ref, path, pathFromRoot, parents, $refs, options) {
   debug('Dereferencing $ref pointer "%s" at %s', $ref.$ref, path);
 
   var $refPath = url.resolve(path, $ref.$ref);
-  var pointer = $refs._resolve($refPath, options);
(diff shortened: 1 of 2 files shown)


realityking:
@BigstickCarpet Could you have a look at this one? Thanks!

BigstickCarpet:
Hi @realityking.  Sorry for taking so long to get to this.

Would you mind adding a test for this?  I just want to make sure that I fully understand the issue, and a test will help with that.  Plus, a test will ensure that I don't accidentally break it again in the future.

kylef:
To note, we're also using the options to disable external references:

```
const swaggerOptions = {
  dereference: {
    circular: 'ignore',
  },
  resolve: {
    external: false,
  },
};
```

jonluca:
AFAICT a refactor in the past changed this behavior. You'll get a `MissingPointerException` if the dereferencing hits a missing reference, and a manual call to `.get` on the reference will just reply with the path the user provided it. Closing for now

PR #90, opened 2018-06-20
fix refs components decoding

RomanGotsiy:
The following ref was incorrectly resolved by the `Pointer.parse` method:

```js
Pointer.parse("#/paths/~1user/application~1vnd.example%2Bjson/schema")
// returns ["paths", "/user", "application/vnd.example%2Bjson", "schema"]
```

Note that `+` in `application/vnd.example%2Bjson` is not decoded.
The reason is usage of `decodeURI`. Instead, `decodeURIComponent` should be used (tokens are components of the pointer)

### DIFF
--- a/lib/pointer.js
+++ b/lib/pointer.js
@@ -167,7 +167,7 @@ Pointer.parse = function (path) {
 
   // Decode each part, according to RFC 6901
   for (var i = 0; i < pointer.length; i++) {
-    pointer[i] = decodeURI(pointer[i].replace(escapedSlash, '/').replace(escapedTilde, '~'));
+    pointer[i] = decodeURIComponent(pointer[i].replace(escapedSlash, '/').replace(escapedTilde, '~'));
   }
 
   if (pointer[0] !== '') {


RomanGotsiy:
Hey @BigstickCarpet 🙌!
Could you please take a look at this?

Thanks in advance!

BigstickCarpet:
Hi @RomanGotsiy - sorry for taking so long to look at this.  I've been super busy at work and haven't had any time.

I'm hesitant to merge this without a test, or at least an example of a schema that doesn't work without this change.

This block of code...
https://github.com/BigstickCarpet/json-schema-ref-parser/blob/9420095f8d717d1ec8372b66a01bbc0d6f903b50/lib/pointer.js#L168-L171

...is directly related to this block of code:
https://github.com/BigstickCarpet/json-schema-ref-parser/blob/9420095f8d717d1ec8372b66a01bbc0d6f903b50/lib/pointer.js#L193-L199

It seems like either _both_ of them should change, or _neither_ of them, but not just one or the other.

RomanGotsiy:
Hey @BigstickCarpet!
No worries! I will add a test a bit later today!

Thanks

RomanGotsiy:
Nice! Thanks a lot for your work! 🙇

PR #103, opened 2018-10-17
saveOriginalRefs option for dereference

RomanGotsiy:
Hey @JamesMessinger,

Here is the PoC of `saveOriginalRefs` option for the `dereference` method.
Before I add docs and tests, could you please check it out?

Also, I'm not sure if the same option makes sense for `bundle`. What do you think?

Thanks!

### DIFF
--- a/lib/dereference.js
+++ b/lib/dereference.js
@@ -4,7 +4,7 @@ var $Ref = require('./ref'),
     Pointer = require('./pointer'),
     ono = require('ono'),
     url = require('./util/url'),
-    setMeta = require('./util/meta').setMeta;
+    setMetadata = require('./util/meta').setMetadata;
 
 module.exports = dereference;
 
@@ -127,7 +127,11 @@ function dereference$Ref ($ref, path, pathFromRoot, parents, $refs, options) {
   }
 
   if (options.dereference.saveOriginalRefs) {
-    setMeta(dereferencedValue, '$ref', $ref.$ref);
+    setMetadata(dereferencedValue, {
+      $ref: $ref.$ref,
+      path: path,
+      pathFromRoot: pathFromRoot,
+    });
   }
 
   return {
--- a/lib/dereference.js
+++ b/lib/dereference.js
@@ -3,7 +3,8 @@
 var $Ref = require('./ref'),
     Pointer = require('./pointer'),
     ono = require('ono'),
-    url = require('./util/url');
+    url = require('./util/url'),
+    setMeta = require('./util/meta').setMeta;
 
 module.exports = dereference;
 
@@ -123,6 +124,10 @@ function dereference$Ref ($ref, path, pathFromRoot, parents, $refs, options) {
     // The pointer is a DIRECT circular reference (i.e. it references itself).
     // So replace the $ref path with the absolute path from the JSON Schema root
     dereferencedValue.$ref = pathFromRoot;
+  }
+
+  if (options.dereference.saveOriginalRefs) {
+    setMeta(dereferencedValue, '$ref', $ref.$ref);
   }
 
   return {
--- a/lib/index.js
+++ b/lib/index.js
@@ -8,11 +8,13 @@ var Options = require('./options'),
     bundle = require('./bundle'),
     dereference = require('./dereference'),
     url = require('./util/url'),
+    getMeta = require('./util/meta').getMeta,
     maybe = require('call-me-maybe'),
     ono = require('ono');
 
 module.exports = $RefParser;
 module.exports.YAML = require('./util/yaml');
+module.exports.getMeta = getMeta;
 
 /**
  * This class parses a JSON schema, builds a map of its JSON references and their resolved values,
--- a/lib/index.js
+++ b/lib/index.js
@@ -8,13 +8,15 @@ var Options = require('./options'),
     bundle = require('./bundle'),
     dereference = require('./dereference'),
     url = require('./util/url'),
-    getMeta = require('./util/meta').getMeta,
+    getMetadata = require('./util/meta').getMetadata,
+    setMetadata = require('./util/meta').setMetadata,
     maybe = require('call-me-maybe'),
     ono = require('ono');
 
 module.exports = $RefParser;
 module.exports.YAML = require('./util/yaml');
-module.exports.getMeta = getMeta;
+module.exports.getMetadata = getMetadata;
+module.exports.setMetdata = setMetadata;
 
 /**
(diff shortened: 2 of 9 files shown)


lib/dereference.js
```diff
@@ -125,6 +126,10 @@ function dereference$Ref ($ref, path, pathFromRoot, parents, $refs, options) {
     dereferencedValue.$ref = pathFromRoot;
   }
 
+  if (options.dereference.saveOriginalRefs) {
+    setMeta(dereferencedValue, '$ref', $ref.$ref);
```
JamesMessinger: Let's add `path` and `pathFromRoot` as well.  I can see those being handy for folks (myself included 😊).   I suggest modifying the `setMetadata()` function to accept multiple properties at once.  That saves us from having to call it multiple times.

```suggestion
    setMetadata(dereferencedValue, {
        $ref: $ref.$ref,
        path: path,
        pathFromRoot: pathFromRoot,
    });
```

lib/index.js
```diff
@@ -8,11 +8,13 @@ var Options = require('./options'),
     bundle = require('./bundle'),
     dereference = require('./dereference'),
     url = require('./util/url'),
+    getMeta = require('./util/meta').getMeta,
     maybe = require('call-me-maybe'),
     ono = require('ono');
 
 module.exports = $RefParser;
 module.exports.YAML = require('./util/yaml');
+module.exports.getMeta = getMeta;
```
JamesMessinger: Let's rename `getMeta()` to `getMetadata()`.  It's more explicit/clear.

```suggestion
module.exports.getMetadata = getMetadata;
```

lib/util/meta.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+var META_SYMBOL = Symbol('json-schema-ref-parser meta');
+
+exports.META_SYMBOL = META_SYMBOL;
```
JamesMessinger: Is there any reason to export this symbol?  I like the idea of it being purely encapsulated in this module, which ensures that people use the `setMetadata()` and `getMetadata()` functions rather than accessing the metadata directly.

lib/util/meta.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Safely sets meta property on the given object value
+ *
+ * @param  {object} obj - A object to set meta info on
+ * @param  {string} prop - Name of property
+ * @param  {string} prop - Value to set
+ */
+exports.setMeta = function (obj, prop, value) {
```
JamesMessinger: Let's rename `setMeta()` to `setMetadata()`. It's more explicit/clear.

```suggestion
exports.setMetadata = function (obj, prop, value) {
```

lib/util/meta.js
```diff
@@ -0,0 +1,33 @@
...
+  obj[META_SYMBOL][prop] = value;
+};
+
+/**
+ * Given the any object returns the associated meta-information or undefined
+ *
+ * @param  {any} value - object to get the meta from
+ * @returns {{pointer: string}}
+ */
+exports.getMeta = function (value) {
```
JamesMessinger: Let's rename `getMeta()` to `getMetadata()`. It's more explicit/clear.

```suggestion
exports.getMetadata = function (value) {
```

RomanGotsiy:
Thanks for reviewing.
I incorporated all your feedback as well as added tests and docs.

JamesMessinger:
Well crap.  I found a couple problems after I merged this PR:

**1. Symbols aren't supported on IE11**
This is an easy fix.  I added code to fallback to an old-fashioned `__metadata` string property if `Symbol` doesn't exist.

**2. Multiple $refs to the same target**
The problem with storing metadata on the dereferenced object itself is that there can be multiple different $refs to the same object, each with different metadata.  I'm experimenting with a couple possible workarounds right now.

JamesMessinger:
I've rolled-back the PR for now and moved the code to [the `feature/metadata` branch](https://github.com/APIDevTools/json-schema-ref-parser/tree/feature/metadata).  We can iterate on it there.

RomanGotsiy:
> 2. Multiple $refs to the same target

Ahh, I didn't consider this as the initial implementation was not adding `pathFromRoot` and `path`. The `$ref` itself shouldn't vary. Right?

JamesMessinger:
Actually, even the `$ref` could be different, since the `$ref` is relative to the file that it's in.  This would most often happen in multi-file schemas, but it could also happen in a single-file one:

**/schema/file1.yaml**
```yaml
type: object
properties:
  firstName:
    $ref: "#/definitions/name"
  middleName:
    $ref: "file1.yaml#/definitions/name"
  lastName:
    $ref: "../schema/file1.yaml#/definitions/name"
  nickName:
    $ref: "file:///schema/file1.yaml#/definitions/name"
  pseudonym:
    $ref: "#/properties/firstName"
definitions:
  name:
    type: string
```

In this example 👆, all five $refs point to the same value, so they'd all be dereferenced to the same object, but each one would have different metadata with a different `$ref`.

I think it doesn't really make sense for the `getMetadata()` function to accept an object (a dereferenced value).  Instead, it should accept a JSON Reference path (e.g. `file1.yaml#/properties/lastName`) and should return metadata for the $ref at that location.   This means that rather than storing metadata as a Symbol on the dereferenced object, it should actually be stored on [the `$Ref` object](https://github.com/APIDevTools/json-schema-ref-parser/blob/master/lib/ref.js).  And the `getMetadata()` method should actually be on [the `$Refs` object](https://github.com/APIDevTools/json-schema-ref-parser/blob/master/lib/refs.js)

kimmobrunfeldt:
We would currently need this feature. The exact use case is that our API documentation generator would need the actual object as well as the name of the Ref. The name will be used as the title for a link in our API docs.

I tested out the current implementation (`github:ReDoc-ly/json-schema-ref-parser#deref-save-original`), but I didn't fully understand the decision to use Symbol. I'm not able to access the metadata even though I have the derefenced object in control. We're using swagger-parser, so as an example I can't do this:

```js
const api = await SwaggerParser.dereference('openapi.yaml', { dereference: { saveOriginalRefs: true }})

// This is not possible, because I'm not sure how the Symbol should be accessed
const metadata = api.paths['/api/v1/myendpoint'].post.requestBody['application/json'].schema.metadata
```

This would a convenient way to access the metadata. Before the saveOriginalRefs is finished, we'll probably implement our own helper which does the dereferencing and injects the metadata of the Ref in the same object.

hypery2k:
@kimmobrunfeldt could you share your approach?

Issue #107, opened 2018-11-22
Feature Request: Proxy support in HTTP resolver


dowlingw:
It would be super helpful to have support for proxies in the HTTP resolver.

Suggest this could be implemented with the following optional fields in the `HTTPResolverOptions` type:
* `proxyHost` - The HTTP proxy host to use (optional)
* `proxyPort` - The port of the HTTP proxy host to use (optional, default 8080)

These could then be passed through to the node http library as the `host` and `port` options in `http.options` respectively when invoked by the HTTP resolver.

Cheers :)

PR #108, opened 2018-11-23
#107: Proxy Support

dowlingw:
Adds support for proxy server and configuration.

Can confirm this approach works using the corporate proxy at my employer.
Other attempts using nodes inbuilt `http`/`https` modules did not work.

Just going to park this here over the weekend to get some feedback from the maintainers.

I'm not sure what testing makes sense, as this quickly gets into testing inside `request` module.

### DIFF
--- a/lib/resolvers/http.js
+++ b/lib/resolvers/http.js
@@ -1,7 +1,6 @@
 "use strict";
 
-var http = require("http"),
-    https = require("https"),
+var request = require("request"),
     ono = require("ono"),
     url = require("../util/url");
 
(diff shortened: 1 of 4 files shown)


JamesMessinger:
Thanks for the PR!  I'll review it soon and get it merged 👍

JamesMessinger:
`request` is a great package for Node.js apps, but `json-schema-ref-parser` is also used in many web apps.  AFAIK, `request` doesn't officially support browser usage, so there's risk of things not working for web apps.  Adding `request` also brings in 48 new dependencies, which will bloat the size of the browser bundle significantly. 

Proxy support is something that would only be needed in Node.js, not web apps (since the browser is responsible for proxy support).  So perhaps there's a way to refactor the [HTTP resolver](https://github.com/APIDevTools/json-schema-ref-parser/blob/master/lib/resolvers/http.js) so that the proxy-related code is only imported by Node.js and not by Browserify.

dowlingw:
Yeah that's fair.

I did note that `withCredentials` isn't explicitly supported by `request` during development, though didn't join the dots to it not supporting browser use entirely.

Will have a noodle over this and see if I can come back with something better, cheers :)

dowlingw:
Closing as I haven't been able to come back to this, sorry! :(

PR #113, opened 2018-12-11
update web browser example in readme

zoellner:
rawgit has shut down. replaced with unpkg


JamesMessinger:
Good catch.  Thanks for the PR!

PR #119, opened 2019-02-21
Allows custom parser declarations

treffynnon:
Adjusts the TypeScript typings to allow custom parsers to be specified.


JamesMessinger:
Thanks!  Very much appreciated

PR #131, opened 2019-07-07
Added ability to customise the set of references that are resolved

stueynz:
Sometimes you don't want all the $ref references to be resolved; added a new options to control which references are dereferenced.  e.g.

```javascript
const options = {
  dereference: {
    circular: true,

    isRefResolved: function(value) {

      // don't resolve where $ref contains the word 'gender'
      return value.$ref.match(/\/gender/g) === null;
    }
  }
```


JamesMessinger:
Seems like this would be a good opportunity to use [file-path-filter](https://github.com/JS-DevTools/file-path-filter), which would let people filter $refs based on globs, regular expressions, custom filter functions, or any combination of them.

Also, it seems like this should be a `resolve` option, not a `dereference` option.  So something like this:

```javascript
const options = {
  resolve: {
    // don't resolve where $ref contains the word 'gender'
    filter: "!**/gender/**"
  }
```

philsturgeon:
That was a cool suggestion James. If anyone would like to pick that up, please make a new PR!

PR #144, opened 2019-12-16
Ignore http and https, add url.

ksaaskil:
- Breaks all browser tests in Karma as they somehow resolve references by downloading schemas. I don't understand how this works, how is it possible that `http` is available in browser tests? Or is `karma` webserver somehow doing the requests on behalf of browser?


ksaaskil:
Sorry I did not mean to make this PR to the base repository, I'm so sorry!

PR #146, opened 2020-01-19
support for nested references

dreisel:
This PR allows for nested references:
for example:
```
{
    "Nats": {
      "$ref": "common://nats",
      "RetryIntervalMs": {
        "$ref": "common://pool:PoolSize"
      }
    }
  }
```

### DIFF
--- a/lib/resolve-external.js
+++ b/lib/resolve-external.js
@@ -57,18 +57,14 @@ function crawl (obj, path, $refs, options) {
     if ($Ref.isExternal$Ref(obj)) {
       promises.push(resolve$Ref(obj, path, $refs, options));
     }
-    else {
-      for (let key of Object.keys(obj)) {
-        let keyPath = Pointer.join(path, key);
-        let value = obj[key];
+    for (let key of Object.keys(obj)) {
+      let keyPath = Pointer.join(path, key);
+      let value = obj[key];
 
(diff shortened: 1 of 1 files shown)


dreisel:
@JamesMessinger What do you think about this change?

JamesMessinger:
Hi @dreisel.  Thank you for your PR. Apologies for the long-delayed response 🙏 

I like this new feature, and I don't see any problems with the code off-hand, but it definitely needs some tests.  I recommend creating a `test/specs/nested/nested.spec.js` file and basically duplicating the contents & structure of the `test/specs/external` folder, but with some nested $refs to test your new functionality.  Be sure to test nested $refs that point to external files as well as nested $refs that point to other locations in the same file.

alexeysuvorov:
Hello, any news on this pull requests? If tests are only concerns I will create atoner pull requests with test. Thanks.

philsturgeon:
@alexeysuvorov hello! 👋  No need for another pull request, you can push tests into this same branch and it'll update this PR. 

Without those tests it's a little tricky for me to see what exactly is going on, which means it's hard to say if I have concerns or not. Can we make sure the change is working as you intended first then go from there?

philsturgeon:
Closing for inactivity. If you can bung some tests into this PR that would help us get this merged.

PR #150, opened 2020-02-05
fix: missing type definition for custom resolver

ayZagen:


JamesMessinger:
Nice!  thank you for submitting that

PR #152, opened 2020-02-20
Include ParserOptions#parse in TS typings

P0lip:
`parse` is missing despite being accepted.
It's not exposed in options in docs and neither is `ResolverOptions#read` that's included in typings.
That being said, it feels like a tiny oversight, hence the PR is here.


JamesMessinger:
Thanks for the PR!  much appreciated

PR #154, opened 2020-02-24
Handle execution under root more correctly

P0lip:
Right now, the file url starts with triple slashes (`///`). This PR fixes it.


lib/util/url.js
```diff
@@ -29,7 +29,7 @@ exports.resolve = require("url").resolve;
  * @returns {string}
  */
 exports.cwd = function cwd () {
-  return process.browser ? location.href : process.cwd() + "/";
+  return process.browser ? location.href : process.cwd() === "/" ? "/" : (process.cwd() + "/");
```
JamesMessinger: This doesn't handle Windows paths.  Let's split this up into two separate steps: 1) get the current CWD, and 2) append a trailing slash if necessary.

```suggestion
  let path = process.browser ? location.href : process.cwd();

  let lastChar = path.slice(-1);
  if (lastChar === "/" || lastChar === "\\") {
    return path;
  }
  else {
    return path + "/";
  }
```

test/specs/absolute-root/absolute-root.spec.js
```diff
@@ -0,0 +1,38 @@
...
+      }
+    };
+  });
+
+  afterEach(() => {
+    url.cwd = cwd;
+    process.cwd = processCwd; // already restored at line 19, but just in case
+  });
+
+  it("should resolve successfully a $ref pointing at absolute location", async () => {
```
JamesMessinger: To match the other test suites, there need to be tests for the `parse()`, `resolve()`, `dereference()`, and `bundle()` methods, each with absolute paths, relative paths, and URLs.  I recommend copying the contents of "test/specs/external" into this directory and using that as your starting point

test/specs/absolute-root/absolute-root.spec.js
```diff
@@ -0,0 +1,38 @@
...
+    let parser = new $RefParser();
+    await parser.parse(path.abs("specs/absolute-root/empty-object.json"));
+    expect(parser.$refs._$refs).to.have.keys([
```
JamesMessinger: Instead of testing against the internal `_$refs` property, you should use the `parser.$refs.paths()` method.  Or, better yet, use the `helper.testResolve()` utility.  See "test/specs/external/external.spec.js" for example usage.

P0lip:
> Can you provide an example of the problem that you're fixing? What does the URL/path that your getting look like currently, and what do you think it should look like?

So, the fs library we use for browsers, operate on `/` as its entry point, therefore `process.cwd()` equals `/`.

test/specs/absolute-root/absolute-root.spec.js
P0lip: > relative paths

This is not going to work since I spoof `process.cwd`. I'd need to spoof fs calls as well.
Are you okay with that?

lib/util/url.js
P0lip: I applied it, but browser is broken. I'm going to eagerly return when `process.browser` is truthy

JamesMessinger:
Sorry for the delayed response.  Last week got crazy busy.

> So, the fs library we use for browsers in case of Stoplight Studio, operates on / as its entry point, therefore `process.cwd()` equals `/`.

Right, but what's the _result_ of this?  What's the problem that you're seeing and fixing?  I want to make sure we're fixing the right problem in the right place.  And that we have the right tests in place.

JamesMessinger:
@P0lip - can you please [enable edits from maintainers](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) on this PR?

P0lip:
> Right, but what's the result of this? What's the problem that you're seeing and fixing? I want to make sure we're fixing the right problem in the right place. And that we have the right tests in place.

The result is that we try reading such paths starting with more than a single slash, i.e `///foo`. Node's `fs` handles them fine, but not every library implementing fs-like interface does it, as the paths are not correct.

> @P0lip - can you please enable edits from maintainers on this PR?

Hm, I cannot seem to find that option.
It's neither visible in the edit view nor in the sidebar :thinking:

JamesMessinger:
> > @P0lip - can you please enable edits from maintainers on this PR?
>
> Hm, I cannot seem to find that option.
> It's neither visible in the edit view nor in the sidebar 🤔

Huh... weird.  I wonder if it's an org-level setting that someone at Stoplight would need to allow. 🤷‍♂️

Anyway... the PR looks good 👍  I have some changes to the tests, which I'll push after merging your changes.

P0lip:
> Huh... weird. I wonder if it's an org-level setting that someone at Stoplight would need to allow. man_shrugging

Hm, that might be the case. I'll touch base with folks to see whether there is such an option exposed somewhere in the panel.

> Anyway... the PR looks good +1 I have some changes to the tests, which I'll push after merging your changes.

Thanks!

PR #163, opened 2020-04-16
Add path information to errors

fmvilas:
Hey! This PR adds extra information to the errors triggered when dereferencing and resolving external references. The reason I'm adding this info is that in the [AsyncAPI playground](https://playground.asyncapi.io), I want to find line and column of the failing $ref and with the current error I'd have to parse the error message, which is not ideal.

I suspect this info could be added to other errors too so feel free to point me out to the other places and I'll be happy to add them on this PR.

Thanks a lot for the excellent tool 🙌

### DIFF
--- a/lib/dereference.js
+++ b/lib/dereference.js
@@ -96,7 +96,13 @@ function dereference$Ref ($ref, path, pathFromRoot, parents, $refs, options) {
   // console.log('Dereferencing $ref pointer "%s" at %s', $ref.$ref, path);
 
   let $refPath = url.resolve(path, $ref.$ref);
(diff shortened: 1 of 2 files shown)


fmvilas:
I think this PR doesn't make sense anymore. #153 also fixed that problem.

BTW, @JamesMessinger, I've seen tests are failing for Windows. Let me know if I can help.

JamesMessinger:
The failing tests on Windows are due to two new dependencies that were added in PR #153 that don't work on IE or Edge.  I opened Issue #165 to discuss possible solutions.

Issue #165, opened 2020-04-20
Stoplight dependencies don't support IE or Edge


JamesMessinger:
@P0lip  and @philsturgeon - I merged PR #153.  All tests pass in Node.js, as well on Chrome, Firefox, and Safari.  But they fail in IE and Edge due to the new dependencies that were added (`@stoplight/json` and `@stoplight/yaml`).  Both of these dependencies use modern ES6+ syntax like rest/spread operators, object destructuring, etc. which is not supported in IE or Edge.

I see a few possible solutions to this problem:

**Option 1: Remove unsupported syntax**
Stoplight could update its libraries to only use syntax that is supported on IE and Edge.  This requires code changes, but no changes are needed to any tooling or build steps. 

**Option 2: Transpile unsupported syntax to supported syntax**
Stoplight could add a transpilation step (e.g. Babel, TypeScript) that converts the unsupported syntax to supported equivalents.  This requires no code changes and allows Stoplight developers to continue using the latest syntax features, but requires changes to the tooling and/or build steps of these libraries.

**Option 3: Drop dependencies**
JSON Schema $Ref Parser could drop the Stoplight dependencies and revert to the previous built-in JSON/YAML parsing logic, which supports IE and Edge.  Stoplight could then replace the built-in parsers with their own parsers in Stoplight projects.

@P0lip  and @philsturgeon  - what are your thoughts on the above options? Is there another option that I haven't considered?

P0lip:
I'd say, we can consider option 3.

The primary reason to have our dependencies is the ability to is more graceful syntax error handling (kind of like fastFail=false for parsing) as well as the ability to produce a range for a particular path.
Now that I think of it, we might not really need it at all? A number of reasons:
- parsing errors are rather rare for the common use cases of json-schema-ref-parser. They are heavily needed in the context of Stoplight (Studio - where users still design specifications, so the documents are not top-notch yet)
- we actually do use ranges

Since I broke it, I'm happy to take that task and switch back to the previous parsers if we decide to go with option 3.

philsturgeon:
3 sounds good, let's get that sorted and I'll review it for ya @JamesMessinger.

PR #166, opened 2020-04-20
Do not use Stoplight packages

P0lip:
Fixes https://github.com/APIDevTools/json-schema-ref-parser/issues/165

Note - I am pretty convinced we should slightly prettify the error messages produced by js-yaml.
Particularly stripping those new lines might beneficial, but I'm fine with keeping them untouched.

I didn't restore YAML tests, not sure if they are truly needed or not - can add them back you find them useful.
Apart from `stringify` method that does not seem to be used anywhere, `parse` is already covered anyway.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -410,6 +410,7 @@ declare namespace $RefParser {
   export class JSONParserError extends Error {
     readonly name: string;
     readonly message: string;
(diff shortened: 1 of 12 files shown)


JamesMessinger:
Thanks for the quick fix @P0lip!  I just published v9.0.0 🎉

Issue #164, opened 2020-04-19
Wrong error.path on dereference


fmvilas:
Hey @P0lip! After the introduction of `failFast`, I'm getting the error `path` property, which is super cool and it's exactly what I need. However, it seems to be pointing to the "wrong" place. Let me explain the context.

I'm trying to make dereference fail on purpose with this document:

```js
{
  "asyncapi": "2.0.0",
  "info": {
    "title": "My API",
    "version": "1.0.0"
  },
  "channels": {
    "mychannel": {
      "publish": {
        "message": {
          "payload": {
            "type": "object",
            "properties": {
              "name": {
                "type": "string"
              }
            }
          }
        }
      }
    }
  }, "components": {
    "messages": {
      "testMessage": {
        "payload": {
          "$ref": "#/components/schemas/testSchema"
        }
      }
    },
    "schemas": {
      "testSchema": {
        "type": "object",
        "properties": {
          "name": {
            "type": "string"
          },
          "test": {
            "$ref": "non-existing.json"
          }
        }
      }
    }
  }
}
```

The failing $ref is at `/components/schemas/testSchema/properties/test/$ref`. However, when I use the following code:

```js
try {
  await $RefParser.dereference(options.source, parsedJSON, {
    parse: options.parse,
    resolve: options.resolve,
    dereference: options.dereference,
    failFast: false,
  });
} catch (e) {
  console.error(e.errors[0].path);
}
```

I get the following `path` property on the error:

```js
[
  'components',
  'messages',
  'testMessage',
  'payload',
  'properties',
  'test'
]
```

Which seems to be the path after resolving local references. Am I missing something? This path is not very useful to me because when I use `getLocationForJsonPath` (from `@stoplight/json` package) I correctly get `undefined`.

Any idea how can we solve this? Thanks!

P0lip:
I see the issue. Agree that we should most likely point at an actual place in the document, rather than the path the resolving process went through.

are you okay with that change @JamesMessinger?

JamesMessinger:
Yep, I think that makes sense.  Go for it!

PR #167, opened 2020-04-22
Throw errors with more meaningful paths

P0lip:
Closes https://github.com/APIDevTools/json-schema-ref-parser/issues/164
I felt like a separate test suite was the best choice here, but happy to spread these new tests across some existing ones if you think that'd be a better option.


JamesMessinger:
@P0lip - It looks like something in this PR doesn't work on Windows.  CI is passing on Linux and Mac but failing on Windows

https://github.com/APIDevTools/json-schema-ref-parser/actions/runs/93093603

P0lip:
@JamesMessinger I see - going to take a look this weekend. Seems like a path might be incorrect (slashes), but I need to take a look to confirm.
Will create a follow-up PR when I have the updated test

JamesMessinger:
Thanks @P0lip !

PR #181, opened 2020-06-05
fix: do not re-crawl unless necessary

marbemac:
Related to the indirect refs bundle optimization: https://github.com/APIDevTools/json-schema-ref-parser/pull/62

For large specifications, the current implementation ends up re-crawling a LOT. So much so that for one customer's spec (which is 48k lines, with inline $refs only, no remote ones), it takes around 20s to bundle. 

Screen Shot 2020-06-05 at 6 47 24 PM

After a brief investigation, it seems like we might not need to re-crawl if we're just updating the inventoried $ref to be a more direct one?

The tests still pass, but I'm not 100% sure on this and so would appreciate some 👀.

For the specification in question, here's the before and after:

Screen Shot 2020-06-05 at 6 47 53 PM

So basically an over 20x increase in performance!

### DIFF
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -135,7 +135,9 @@ function inventory$Ref ($refParent, $refKey, path, pathFromRoot, indirections, i
   });
 
   // Recursively crawl the resolved value
-  crawl(pointer.value, null, pointer.path, pathFromRoot, indirections + 1, inventory, $refs, options);
+  if (!existingEntry) {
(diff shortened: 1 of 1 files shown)


philsturgeon:
@JamesMessinger congratulations mate!

marbemac:
> I just had a baby, so I've been AFK for a few weeks. 👶🍼

Congrats! Very happy for you, hope all is well :).

JamesMessinger:
Thanks guys!  BTW... This PR has now been released in version 9.0.2

PR #186, opened 2020-06-17
Fixes treating empty parsing results

karol-maciaszek:
This PR fixes the condition of rejecting empty parsing results.

Currently, when `options.continueOnError` is true and result is empty it will reject.
With this change, it will reject only if `allowEmpty` is false. 

I'm a bit unsure whether this should be still dependent on `continueOnError`.


JamesMessinger:
Requesting @P0lip to take a look at this, since he knows that logic best

Issue #185, opened 2020-06-17
Omitting TypedArrays from crawling?


ntamas:
I have recently run into an issue with `json-schema-ref-parser` on a JSON object that uses `$ref` pointers to binary assets that get merged into the JSON object as `Uint8Array`s. Some of these arrays are fairly large (several megabytes), and it turns out that the `crawl()` function in `lib/dereference.js` descends into every single element of these arrays even though they cannot contain `$ref` references to other items (since all the elements are numbers).

Would it be possible to exclude `TypedArray`s from the crawl? If so, would you be open to accepting a PR if I prepare one?

JamesMessinger:
Yeah, I think that makes total sense. A PR would be appreciated 👍

philsturgeon:
Fixed in #187.

PR #187, opened 2020-06-24
Skip typed arrays and DataViews while crawling the schema

ntamas:
This PR addresses the issue outlined in #185: when a schema contains typed arrays or DataViews, this part can safely be skipped during crawling because it cannot contain references.

This PR is marked as a draft because I couldn't figure out how to add tests for this. Initially I wanted to solve this with `chai-spies` such that a spy logs all the paths that the crawler crawled and checks whether the items of typed arrays are excluded from the crawl, but I could not add a spy on the crawler function as it is internal. Any help would be appreciated.

All the existing tests pass, except the HTTP-based ones which fail for a reason unrelated to the PR. Feel free to merge if you think that it is not required to add test cases to cover the changes in this PR.

### DIFF
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -39,7 +39,7 @@ function bundle (parser, options) {
 function crawl (parent, key, path, pathFromRoot, indirections, inventory, $refs, options) {
   let obj = key === null ? parent : parent[key];
 
-  if (obj && typeof obj === "object") {
+  if (obj && typeof obj === "object" && !ArrayBuffer.isView(obj)) {
     if ($Ref.isAllowed$Ref(obj)) {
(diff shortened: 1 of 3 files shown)


JamesMessinger:
@ntamas - Thanks for this PR!  This is a great performance tweak, and it taught me about the `ArrayBuffer.isView()` method, which I didn't know about. 

I don't think you need to add any specific tests for this, since there are already plenty of tests that involve Buffers and DataViews, which will exercise your new code.  I think the PR is good to merge as-is.  Can you remove the "draft" status, so GitHub will let me merge it?

JamesMessinger:
@ntamas - Nevermind!  I figured out how to merge the PR even in "draft" status. ;)   Thanks again!

ntamas:
Awesome, thanks a lot for merging!

PR #189, opened 2020-07-17
Add path token to resolved pointer value

nkappler:
When dereferencing a json schema, the `#/definitions` are merged into the root type.
As a result, when using json-schema-to-typescript to generate declaration files from the schema,
all externally referenced interfaces get merged into one big interface.
See: https://github.com/bcherny/json-schema-to-typescript/issues/143

By sticking the token to the resolved pointer value, the information about the path (i.e. the interface name) is not lost
and can be used to generate the right interfaces. The corresponding change can be found here:
https://github.com/nkappler/json-schema-to-typescript/commit/d53af1210d1514cf5b3e03723e4a24d02f065e73

*I am not sure if this is the 'proper' way to do it, but it is a tiny and simple fix and it works*

### DIFF
--- a/lib/pointer.js
+++ b/lib/pointer.js
@@ -73,6 +73,7 @@ function Pointer ($ref, path, friendlyPath) {
  */
 Pointer.prototype.resolve = function (obj, options) {
   let tokens = Pointer.parse(this.path, this.originalPath);
+  let token;
 
   // Crawl the object, one token at a time
   this.value = unwrapOrThrow(obj);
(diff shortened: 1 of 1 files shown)


JamesMessinger:
Thanks for the PR!  I'm having trouble seeing what problem this solves and/or what effect it will have for consumers of the library.  Could you add some tests that exercise the behavior and only pass with this PR in place?

nkappler:
Hi James,
the problem is described in detail here: bcherny/json-schema-to-typescript#143
Also see my comments there from yesterday, where i found the root cause

nkappler:
Especially this comment:
https://github.com/bcherny/json-schema-to-typescript/issues/143#issuecomment-659349106

philsturgeon:
Unclear problem being solved. Sorry!

PR #190, opened 2020-07-23
Do not mutate objects when values do not change

haggholm:
I had some code break in unexpected ways when ref-parser mutated my schemata (my fault—it doesn’t say it won’t), and deep-froze my inputs to catch it. I then discovered that ref-parser sometimes performs mutations on objects *even when the values do not change*. I suppose frozen schema objects may not be a super common use case, but it seems like a silly thing to break over: the equivalent of `obj[key] = obj[key]` after all does nothing *except* to break if `obj` is frozen.

### DIFF
--- a/lib/dereference.js
+++ b/lib/dereference.js
@@ -57,13 +57,19 @@ function crawl (obj, path, pathFromRoot, parents, $refs, options) {
         if ($Ref.isAllowed$Ref(value, options)) {
           dereferenced = dereference$Ref(value, keyPath, keyPathFromRoot, parents, $refs, options);
           circular = dereferenced.circular;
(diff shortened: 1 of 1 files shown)


JamesMessinger:
I agree that mutating the schema is problematic, but it's difficult to change that behavior at this point.  But at the very least, we shouldn't attempt _unnecessary_ mutations.  So thanks for this PR!

jmm:
I was coming here to report that this mutates the input schema (e.g. when calling `bundle()`).

It would be better for it to not mutate the input, but if it's going to it would at least be better to document that.

@JamesMessinger why would it be difficult to change?

PR #195, opened 2020-10-20
Performances problems fixes on big files with a lot of references on same big object

### LINKED ISSUE #176: Performance issue
First of all, thanks for writing this library :)

We're running into a performance issue that I wanted to alert you to.

We're loading a fairly large (~900KB) JSON schema file, parsing it, and running the resulting JavaScript object through this library's `dereference()` function.  The `dereference()` call takes around 68 seconds to complete.

We've disabled all loading of local or external files, so I don't think that any of the time spent in `dereference()` is due to loading external resources.

Here's basically what the code looks like:

```JavaScript
const rawSchema = fetch(...);
const schema = JSON.parse(rawSchema);

if (typeof (schema) !== 'object')
	throw `The schema was not an object.`;

const options = {
	parse: {
		json: false,
		yaml: false,
		text: false,
		binary: false
	},
	resolve: {
		external: false,
		file: false,
		http: false
	},
	continueOnError: false,
	dereference: {
		circular: true
	}
};

console.log(performance.now());

$SchemaRefParser
	.dereference(schema, options)
	.then(dereferencedSchema => {
		console.log(performance.now());
	});
```

The schema does contain circular references.

paztis:
fixes #176.

add processedObject list to avoid recursion in already processed tree parts
add dereferencedCache to avoid to  re-treat object already dereferenced

### DIFF
--- a/lib/dereference.js
+++ b/lib/dereference.js
@@ -16,7 +16,7 @@ module.exports = dereference;
  */
 function dereference (parser, options) {
   // console.log('Dereferencing $ref pointers in %s', parser.$refs._root$Ref.path);
-  let dereferenced = crawl(parser.schema, parser.$refs._root$Ref.path, "#", [], parser.$refs, options);
+  let dereferenced = crawl(parser.schema, parser.$refs._root$Ref.path, "#", [], [], {}, parser.$refs, options);
   parser.$refs.circular = dereferenced.circular;
   parser.schema = dereferenced.value;
 }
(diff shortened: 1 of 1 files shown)


paztis:
I was working on a file that takes 10s to be processed (11 millions of passes in the crawl function)
I add caches to avoid this, and now we only pas 35000 times inside

marbemac:
Very nice - just curious, could the `bundle` method benefit from a similar caching mechanism?

paztis:
I don't look at the bundle usage. I only use this one, but potentially yes.

paztis:
the same dev will be to redo

paztis:
@P0lip any updates on the review ?

marbemac:
He's on vacation until early November, we'll have time to check it out then!

P0lip:
Hey @paztis.
As Marc stated, I'm off until the end of this week.
I'll try to get back to you on Monday, next week (02 Nov)

Bosch-Eli-Black:
> Hey @paztis.
> As Marc stated, I'm off until the end of this week.
> I'll try to get back to you on Monday, next week (02 Nov)

Enjoy your vacation 🙂

lib/dereference.js
```diff
@@ -28,60 +28,65 @@ function dereference (parser, options) {
...
+function crawl (obj, path, pathFromRoot, parents, processedObjects, dereferencedCache, $refs, options) {
   let dereferenced;
   let result = {
     value: obj,
     circular: false
   };
 
-  if (obj && typeof obj === "object" && !ArrayBuffer.isView(obj)) {
-    parents.push(obj);
+  if (options.dereference.circular === "ignore" || processedObjects.indexOf(obj) === -1) {
```
P0lip, line 44: Was `options.dereference.circular` added here accidentally?
I don't think this should be placed here, as this will likely prevent all objects from being crawled.
Let's add a test for it too. We don't seem to have it covered.

The if statement could be combined as well, but that's not a must. It's just a nitpick.

lib/dereference.js
```diff
@@ -94,14 +99,37 @@ function crawl (obj, path, pathFromRoot, parents, $refs, options) {
...
+    if (refKeys.length > 1) {
+      const extraKeys = {};
+      for (let key of refKeys) {
+        if (key !== "$ref" && !(key in cache.value)) {
+          extraKeys[key] = $ref[key];
+        }
+      }
+      return {
+        circular: cache.circular,
+        value: Object.assign({}, cache.value, extraKeys),
```
P0lip, line 125: This might screw the order of properties and override some of them.
I'd suggest we bail out to regular scenario, same as we do in case of storing the cache at line 176.
Alternatively, we could maintain 'before$RefKeys' and `after$RefKeys` and merge them in that order.

lib/dereference.js
```diff
@@ -94,14 +99,37 @@ function crawl (obj, path, pathFromRoot, parents, $refs, options) {
  * @param {string} path - The full path of `$ref`, possibly with a JSON Pointer in the hash
  * @param {string} pathFromRoot - The path of `$ref` from the schema root
  * @param {object[]} parents - An array of the parent objects that have already been dereferenced
+ * @param {object[]} processedObjects - An array of all the objects that have already been dereferenced
```
P0lip, line 102: Missing jsdoc annotation for `dereferencedCache`
```suggestion
 * @param {object[]} processedObjects - An array of all the objects that have already been dereferenced
 * @param {object} dereferencedCache - An map of all the dereferenced objects 
```

P0lip (commented):
Would be great to add some tests for the properties merging (line 125).
Looking pretty solid otherwise

lib/dereference.js
paztis: in the existing unit tests, it was failing without.
In fact in the process you've got, you dereference the object, then you re-reference if it circular === "ignore"
With my cache it was not execing the re-reference code inside the loop, that's why I add this condition

lib/dereference.js
paztis: yes it will set extraKeys after basic ones
Is this order important ?

I do it to support some unit tests where an object has a description, then a reference of it overrides the description.
The cache key is the same for both (reference is unique). But after dewreference, ths description in both versions is not the same.
You've got specific unit tests for this.

P0lip: >  Is this order important ?

Yeah it might matter at times if you have the property with the same key - it'll get overriden, won't it?
IMHO the current implementation may have a minor bug here.
I went through tests and we don't seem to have anything that'd validate the scenario I mentioned.

The below scenario fails
```js
  it("should preserve the order of properties", async () => {
    let parser = new $RefParser();
    const schema = await parser.dereference(
      // This file doesn't actually need to exist. But its path will be used to resolve external $refs
      path.abs("path/that/does/not/exist.yaml"),
      // This schema object does not contain any external $refs
      {
        foo: {
          minLength: 1,
        },
        bar: {
          $ref: "#/foo",
        },
        baz: {
          $ref: "#/foo",
          minLength: 4,
        },
      },
      // An options object MUST be passed (even if it's empty)
      {});

    expect(schema).to.deep.equal({
      foo: {
        minLength: 1
      },
      bar: {
        minLength: 1
      },
      baz: {
        minLength: 4
      },
    });
  });
```

An improved implementation could look as follows
```js
  if (refKeys.length > 1) {
      const value = {};
      for (let key of refKeys) {
        if (key === "$ref") {
          Object.assign(value, cache.value);
        }
        else {
          value[key] = $ref[key];
        }
      }

      return {
        circular: cache.circular,
        value,
      };
    }
```

This is questionable, however, since, if I am not mistaken, until draft 2019-09 you couldn't place $refs to any other properties, but as of 2019-09 you can and IIRC they are merged similarly to YAML merge keys.
I took at look at the [spec](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.8.2.4), but couldn't really find any information regarding the order of merging.
Perhaps @philsturgeon has a clue on this one?

lib/dereference.js
P0lip: Oh, I misread the condition.
My apologies.
`if (options.dereference.circular !== "ignore" && processedObjects.indexOf(obj) === -1) { `
For whatever reason, I thought of it as above ^.

steverice:
FWIW this resolves the same issue that I have with an OpenAPI schema containing many (hundreds) of refs.
@P0lip @philsturgeon is this on track to be merged soon?

P0lip:
@steverice there is one (minor) suggestion left to be addressed.
I think we can merge it as is and I could apply the comment later on. It's non-blocking.

tahpot:
Thanks to all for the work on this. Fantastic!

Bosch-Eli-Black:
@P0lip Can we get a new release on NPM? 🙂  (Beta would be fine)

PR #197, opened 2020-11-09
add stripQuery method and strip the query from the extension

mathieudutour:
Given a url `https://example.com/file.yml?foo=bar`, the extension will be parsed as `.yml?foo=bar` and the relevant parsers will miss it.

This PR fixes this issue by stripping the query from the extension.


P0lip (commented):
The code looks good.
Would you mind adding a small test for it?

mathieudutour:
done 👍

philsturgeon (approved):

philsturgeon:
Thank you @mathieudutour!

Issue #29, opened 2016-11-22
Overriding port for absolute refs when run from browser [3.1.2]


andru:
I've got a schema with absolute urls to resources with no port number in the URL...

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "id": "http://localhost/schema/01/plantdb.json",
  "type": "object",
  "properties": {
    "names": {"$ref": "http://localhost/schema/01/names.json"}
  }
}
```
(greatly simplified for clarity)

and I'm loading this on the client from an app hosted at `localhost:3000`. Trying to dereference a schema file hosted on the same hostname, but without specifying a port number...

```js
parser.dereference('http://localhost/schema/01/plantdb.json')
``` 

actually tries to load the schema from `http://localhost:3000/schema/01/plantdb.json` and fails with an invalid schema error. If I explicitly add the port number to the schema URL:

```js
parser.dereference('http://localhost:80/schema/01/plantdb.json')
```

then that schema file loads, but all refs within are still loaded from port `http://localhost:3000`.

Is this a bug, or is there a config somewhere I'm missing to force the port number on ref requests?

BigstickCarpet:
There's nothing in JSON Schema $Ref Parser that would add port 3000 to URLs.  You can verify this yourself by checking the URL at [this line of code](https://github.com/BigstickCarpet/json-schema-ref-parser/blob/master/lib/resolvers/http.js#L104) to see the HTTP request that is being sent.

Perhaps the server you're using is doing some redirect or proxying behavior?

andru:
I've done some debugging and I _think_ there's a bug here.

This script is running on a page served from localhost:3000.

```js
parser.dereference('http://localhost/schema/01/plantdb.json')
```
_quick sidenote: A simple fix here is appending :80 to the hostname, but then the bug just moves to the request for each ref URL within (which don't have port numbers, assuming the sensible default of 80 for http) which I ordinarily wouldn't have control over, so I'm keeping it simple here by leaving the port off the initial request_

This URL gets parsed to the following object as a function parameter to `get` in `lib/resolvers/http.js`

```json
{
  "protocol":"http:",
  "slashes":true,
   "auth":null,
  "host":"localhost",
  "port":null,
  "hostname":"localhost",
  "hash":null,
  "search":null,
  "query":null,
  "pathname":"/schema/01/plantdb.json",
  "path":"/schema/01/plantdb.json",
[...]

PR #198, opened 2020-12-04
fix: request the schema id URL instead of the current page url

mmcorrelo:
**When the schema id is defined with an URL should use that hostname to request the references, instead of using the current page URL**

EXAMPLE: 

```json
{
	"$schema": "http://json-schema.org/draft-07/schema#",
	"$id": "https://resourse.example.com/schemas/resources/main.json#",
	"title": "This is my main schema",
	"description": "This is my main schema description.",
        "type": "object",
	"properties": {
	  "$ref": "/schemas/resources/version1.0.1.json#"
	}
  }
```

Currently, if the dereference work is being done locally at http://localhost:3000 the deref request will be done with URL  http://localhost:3000/schemas/resources/version.1.0.1.json instead of https://resourse.example.com/schemas/resources/version1.0.1.json

This is related to this [issue]( https://github.com/APIDevTools/json-schema-ref-parser/issues/29)


philsturgeon:
@Relequestual hey could I bother you for a quick check here? The issue is outlined in #199 and we want to make sure the case is even valid before merging this.

PR #202, opened 2021-01-05
use isomorphic-fetch simplify http resolver

septs:
fix https://github.com/APIDevTools/json-schema-ref-parser/issues/107


philsturgeon:
Conceptually this seems like a good idea, but a few things:

Firstly, is there any documentation required to fully solve #107, does the user need to change anything or is it just magic?

Secondly. another attempt was made to fix #107 by using request in #108, and that was closed as it doesn't work in the browser. fetch() does work in the browser ofc, but we might need to do a little more work to make sure node-fetch is only loaded in node and normal fetch is available in the browser. 

@P0lip I'd appreciate your input here on the second thing, because this will effect @stoplightio stuff, probably for the better.

septs:
> Secondly. another attempt was made to fix #107 by using request in #108, and that was closed as it doesn't work in the browser. fetch() does work in the browser ofc, but we might need to do a little more work to make sure node-fetch is only loaded in node and normal fetch is available in the browser.

support via `browser` field in package.json on browser downgrade to native fetch

see https://github.com/node-fetch/node-fetch/blob/2.x/package.json#L6
see https://github.com/node-fetch/node-fetch/blob/2.x/package.json#L6
see https://github.com/defunctzombie/package-browser-field-spec

package.json
```diff
@@ -72,6 +72,7 @@
   "dependencies": {
     "@jsdevtools/ono": "^7.1.3",
     "call-me-maybe": "^1.0.1",
-    "js-yaml": "^3.13.1"
+    "js-yaml": "^3.13.1",
+    "node-fetch": "^2.6.1"
```
P0lip, line 76: I'm uncertain about this package, as this is clearly meant to be used in the context of Node.js, while json-schema-ref-parser is also supposed to work in modern browsers.
Although I know we used `http` and `https` Node modules in the existing codebase, I suggest we use [isomorphic-fetch](https://www.npmjs.com/package/isomorphic-fetch).
Most of robust bundlers such as Webpack or Rollup were able to shim Node moduels such as `http` and `https` and doing so might be potentially slightly more tricky if we rely on external libraries (that may change their implementation any time).

P0lip (changes requested):
Thanks for the PR!
I'm very supportive when it comes to getting rid of `http` and `https`. It's a step in good direction since this makes the integration of json-schema-ref-parser in browser environment much easier.
There are a few comments left to address, however.
Let me know if anything needs further clarification. Happy to help getting this PR over the finish line.

lib/resolvers/http.js
```diff
@@ -58,9 +95,16 @@ module.exports = {
...
     }
 
-    return fetch(this.makeRequest(url.format(u)))
-      .then((response) => response.arrayBuffer())
-      .then((buffer) => Buffer.from(buffer))
+    const request = new Request(url.format(u), {
+      method: "GET",
+      headers: this.headers,
+      redirect: this.redirects,
+      timeout: this.timeout,
```
P0lip, line 93: Unfortunately, `Request` does not support timeout.
You need to use [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController).
There is a neat usage example included in the link above.

lib/resolvers/http.js
```diff
@@ -58,9 +95,16 @@ module.exports = {
       u.protocol = url.parse(location.href).protocol;
     }
 
-    return fetch(this.makeRequest(url.format(u)))
-      .then((response) => response.arrayBuffer())
-      .then((buffer) => Buffer.from(buffer))
+    const request = new Request(url.format(u), {
+      method: "GET",
+      headers: this.headers,
+      redirect: this.redirects,
```
P0lip, line 92: I suggest we drop it.
[`redirect`](https://developer.mozilla.org/en-US/docs/Web/API/Request/redirect) is an entirely different setting in case of `Request`.

lib/resolvers/http.js
```diff
@@ -80,101 +85,17 @@ module.exports = {
...
 
-    return download(u, this);
-  }
+    /** @type {RequestInit} */
+    const init = {
+      method: "GET",
+      headers: this.headers,
+      redirect: this.redirects,
+      timeout: this.timeout,
+      credentials: this.withCredentials ? "include" : undefined,
```
P0lip, line 94:
```suggestion
      credentials: this.withCredentials ? "include" : "omit",
```

lib/resolvers/http.js
P0lip:
```suggestion
```

lib/index.d.ts
```diff
@@ -255,6 +255,11 @@ declare namespace $RefParser {
      * Set this to `true` if you're downloading files from a CORS-enabled server that requires authentication
      */
     withCredentials?: boolean;
+
+    /**
+     * Provide `fetch` function
+     */
+    fetch: typeof globalThis["fetch"];
```
P0lip, line 262: Let's add to the top of the file.
We cannot assume the consumer will have the `dom` lib included in their TS config.

```suggestion
    // eslint-disable-next-line
   /// <reference lib="dom" />
```

package-lock.json
```diff
@@ -1,8 +1,14918 @@
 {
   "name": "@apidevtools/json-schema-ref-parser",
   "version": "9.0.6",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
```
P0lip, line 7: Hmmm. Something odd happened here.
Could you execute this?
```
git checkout 00181a7~1  -- package-lock.json
npm install
```

lib/resolvers/http.js
septs: see https://github.com/node-fetch/node-fetch#request-cancellation-with-abortsignal

i think, remove timeout option may be better.

septs: if so, need to add extra library, https://npm.im/abort-controller

P0lip: Let's add the library. The fact the last release took place 2 years is nothing wrong. The spec has simply not changed since then, so there was no need to introduce any change.
I had a chance to utilize that library in the past and I can say it's solid.

philsturgeon:
I think removing timeouts and redirect logic is a little nuclear here, as there are alternative approaches.

To get timeouts working with abort controller: https://dmitripavlutin.com/timeout-fetch-request/

There's some stuff on google about follow redirects being built in https://stackoverflow.com/questions/39735496/redirect-after-a-fetch-post-call#39739894 but im not sure if thats going to work for this specific implementation.

Anyhow, if we remove these that's probably a major release, and it's not a great sales pitch to just say there's less functionality now.

philsturgeon:
Btw I appreciate your work on this! Sorry if that doesn't come across, I write a little too quickly sometimes, never mean to sound rude! 😅

septs:
i planning use typescript design a new library.

philsturgeon:
@septs I don't understand. You had almost finished this feature and instead something something now you're doing a new one in TypeScript? 

There's already a TypeScript rewrite of this library that's coming along quite nicely, and it does a lot more than just switch to TS. There's years of domain knowledge built up in this codebase and that one. Would you like to work together instead of duplicating efforts?

PR #204, opened 2021-01-15
feat(http-resolver): use isomorphic-fetch for making requests

### LINKED ISSUE #107: Feature Request: Proxy support in HTTP resolver

P0lip:
Fixes #107.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -119,7 +119,7 @@ $RefParser.prototype.parse = async function parse (path, schema, options, callba
   try {
     let result = await promise;
 
-    if (result !== null && typeof result === "object" && !Buffer.isBuffer(result)) {
+    if (result !== null && typeof result === "object" && !ArrayBuffer.isView(result)) {
       me.schema = result;
       return maybe(args.callback, Promise.resolve(me.schema));
     }
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -160,5 +160,5 @@ function isEmpty (value) {
   return value === undefined ||
     (typeof value === "object" && Object.keys(value).length === 0) ||
     (typeof value === "string" && value.trim().length === 0) ||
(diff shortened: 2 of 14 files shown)


lib/dereference.js
```diff
@@ -100,7 +100,7 @@ function crawl (obj, path, pathFromRoot, parents, processedObjects, dereferenced
  * @param {string} pathFromRoot - The path of `$ref` from the schema root
  * @param {object[]} parents - An array of the parent objects that have already been dereferenced
  * @param {object[]} processedObjects - An array of all the objects that have already been dereferenced
- * @param {object} dereferencedCache - An map of all the dereferenced objects 
+ * @param {object} dereferencedCache - An map of all the dereferenced objects
```
philsturgeon, line 103: No need to review the rest of this PR, I'm in!

philsturgeon:
@P0lip if you base this on the `beta` branch we can see if it works! It looks good so far, but when you're ready lets see if we can make a 9.2.0-beta. :) Don't forget to make a commit message of something like `feat(resolvers): HTTP requests now support Proxy`.

We won't always need to make a beta, just a test! :)

philsturgeon:
Re-implements https://github.com/APIDevTools/json-schema-ref-parser/pull/202

philsturgeon (approved):

philsturgeon:
Awesome, merge away!

PR #207, opened 2021-01-18
Semantic Release

philsturgeon:
Semantic Release is a mixed bag. Some people love it, some people hate it, I try and take a more reasoned approach and use it where it fits, and I think it fits this project. It's great for hands off releasing on projects where folks are pretty damn busy and manual admin is more of a hastle than its worth. This package was a bit out of sync with itself, latest "GitHub Release" was 2.2.0, latest tag was 9.0.6, latest CHANGELOG.md was 9.0.0, and latest NPM was v9.0.7. 

Instead of sending PRs to change the version in `package.json` we can just merge PRs. The [commit format](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines) looks like this:

Commit message | Release type
-- | --
fix(pencil): stop graphite breaking when too much pressure applied | Patch Release
feat(pencil): add 'graphiteWidth' option | Minor Feature Release
perf(pencil): remove graphiteWidth optionBREAKING CHANGE: The graphiteWidth option has been removed.The default graphite width of 10mm is always used for performance reasons. | Major Breaking Release

We can update commit messages on behalf of users when we merge in order to avoid blocking PRs or confusing contributors. It'll then publish to NPM, create a GitHub Releases (which will also tag) and update CHANGELOG.md.

It _should_ work ok, and merging to beta will show how it works. If you like it, we can put it in master. If it fails, I'll fix it. 😆

### DIFF
--- a/.github/workflows/CI-CD.yaml
+++ b/.github/workflows/CI-CD.yaml
@@ -117,7 +117,8 @@ jobs:
 
   release:
     name: Release
-    if: github.ref == 'refs/heads/master'
+    # removed to test beta branch.
+    # if: github.ref == 'refs/heads/master'
     runs-on: ubuntu-latest
     timeout-minutes: 10
     needs:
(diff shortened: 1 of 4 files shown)


JamesMessinger (approved):
Sweet!  Thanks for taking this on, Phil!

P0lip (approved):
Looks :ok: at first glance.
We could probably add https://github.com/semantic-release/commit-analyzer to the mix, but can skip it for now.

PR #228, opened 2021-04-19
external refs: keep track of original paths during resolve

paulRbr:
Right now the library is “loosing” the original value
information from the `$ref` entry in case of we are resolving relative
file paths.

However, it can be a useful information to keep. This PR is thus trying
to keep a `Set` containing all original paths when resolving external
$ref pointers.


jonluca:
I'm closing stale PRs - happy to merge in a new PR if it's rebased

PR #231, opened 2021-05-16
stop infinite recursion w/ crawling external refs

cody-greene:
This will fix a few cases of "Uncaught RangeError: Maximum call stack size exceeded"

Specifically if you have a schema with a circular reference like this:
```javascript
const Node = {
  type: 'object',
  properties: {
    child: {$ref: '#'}
  }
}
```
And for whatever reason this gets dereferenced twice:
```javascript
// First call runs fine.
const dereferenced = await $RefParser.dereference(Node)

console.log(dereferenced)
//{
//  type: 'object',
//  properties: { child: [Circular *1] }
//}

// Infinite recursion will occur in resolve-external.js:crawl()
await $RefParser.dereference(dereferenced)
```

And of course this can happen without calling `dereference()` twice. In fact, I can construct a dereferenced schema myself like so:
```javascript
const Node = {
  type: 'object',
  properties: {
    child: null  
  }
}
Node.properties.child = Node

// Infinite recursion will occur 
await $RefParser.resolve(Node)
```

To resolve this we can modify the recursive function to track seen objects and return early if one is seen second time (or more).

### DIFF
--- a/lib/resolve-external.js
+++ b/lib/resolve-external.js
@@ -44,6 +44,7 @@ function resolveExternal (parser, options) {
  * @param {string} path - The full path of `obj`, possibly with a JSON Pointer in the hash
  * @param {$Refs} $refs
  * @param {$RefParserOptions} options
+ * @param {WeakMap} seen - Internal.
  *
  * @returns {Promise[]}
  * Returns an array of promises. There will be one promise for each JSON reference in `obj`.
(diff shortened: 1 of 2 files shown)


lib/resolve-external.js
```diff
@@ -44,17 +44,20 @@ function resolveExternal (parser, options) {
...
  *
  * @returns {Promise[]}
  * Returns an array of promises. There will be one promise for each JSON reference in `obj`.
  * If `obj` does not contain any JSON references, then the array will be empty.
  * If any of the JSON references point to files that contain additional JSON references,
  * then the corresponding promise will internally reference an array of promises.
  */
-function crawl (obj, path, $refs, options) {
+function crawl (obj, path, $refs, options, seen) {
+  seen = seen || new WeakMap();
```
P0lip, line 56: Any reason not to use plain Set?

P0lip (commented):
Looks good!

lib/resolve-external.js
cody-greene: A plain Set makes more sense, for sure.  Here I go rolling my eyes at my own code again :roll_eyes:

philsturgeon (approved):

PR #234, opened 2021-05-25
refs: resolve file system path correctly with encoded values

paulRbr:
The internal `_$refs` object stores all references path with an URL
encoded value.

However when using the public API of the $refs object
most user will, most probably, expect to pass the decoded file
path to all public methods (https://github.com/APIDevTools/json-schema-ref-parser/blob/master/docs/refs.md#methods)

This commit makes sure to search for URL encoded path value when the
path is a file system path.

### DIFF
--- a/lib/refs.js
+++ b/lib/refs.js
@@ -148,6 +148,9 @@ $Refs.prototype._resolve = function (path, pathFromRoot, options) {
   let absPath = url.resolve(this._root$Ref.path, path);
   let withoutHash = url.stripHash(absPath);
   let $ref = this._$refs[withoutHash];
+  if (url.isFileSystemPath(withoutHash)) {
+    $ref = this._$refs[url.fromFileSystemPath(withoutHash)];
+  }
 
   if (!$ref) {
     throw ono(`Error resolving $ref pointer "${path}". \n"${withoutHash}" not found.`);


philsturgeon (changes requested):
Can you add some tests to show the behavior is working as expected?

paulRbr:
Thanks for pointing that to me @philsturgeon I should have done that from the start. I think this patch is incorrect and I misunderstood the library internals.

Could you help me out?

Basically, I thought I could transparently use the resulting paths from the [`$Refs.paths()` function](https://apitools.dev/json-schema-ref-parser/docs/refs.html#pathstypes) as inputs to the [`$Refs.get($ref)` function](https://apitools.dev/json-schema-ref-parser/docs/refs.html#getref). But the documentation is clear: the `get($ref)` function takes a “JSON Reference path, optionally with a JSON Pointer in the hash” and not a path/URL as I'm trying to do.

Maybe another solution would be to allow a new optional parameter to the `$Refs.paths()` function which would return **encoded file path list** (instead of the **current decoded path list**).

What do you think?

paulRbr:
I've pushed my new suggestion [here](https://github.com/paulRbr/json-schema-ref-parser/commit/94fc5ee26bb470d9cd846e183627fb3c598e8b2d) before I get another feedback :)

paulRbr:
I will close this PR for now as I guess I was trying to bypass the whole point of the existing `$Refs.values()` function (that I missed). By returning an object with decoded paths as keys, this function is exactly what you should use when searching references from their original filesystem paths. Sorry @philsturgeon for the confusion!

Issue #201, opened 2020-12-17
Root ref fails to parse


isaacl:
```json
{
  "$ref": "#/definitions/EmailConfig",
  "$schema": "http://json-schema.org/draft-07/schema#",
  "definitions": {
    "EmailConfig": {
      "additionalProperties": false,
      "properties": {
        "adminContact": {
          "type": "string"
        }
      }
    }
  }
}
```
is valid (in fact is part of the [test suite here](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/2019-09/tests/draft7/ref.json#L122-L129)).  But json-schema-to-typescript fails to resolve the ref using this library.

---------
Originally filed at bcherny/json-schema-to-typescript#349

vinyasns:
@isaacl Yes I faced the same issue, it gives 'Circular $ref pointer found ...' error even though this is a valid case.

philsturgeon:
If this is still an issue please can somebody send a pull request.

PR #246, opened 2021-09-23
fix: root level reference internally do not resolved 

jonaslagoni:
This PR addresses the issue where root level references to internal schema definitions are not resolved accurately.

Fixes https://github.com/APIDevTools/json-schema-ref-parser/issues/201

### DIFF
--- a/test/specs/root-internal/bundled.js
+++ b/test/specs/root-internal/bundled.js
@@ -1,0 +1,19 @@
+"use strict";
+
+module.exports =
+{
+  title: "name",
+  required: [
+    "first",
+    "last"
+  ],
+  type: "object",
+  properties: {
+    last: {
+      type: "string"
+    },
+    first: {
+      type: "string"
+    }
+  },
+};
--- a/test/specs/root-internal/dereferenced.js
+++ b/test/specs/root-internal/dereferenced.js
@@ -1,0 +1,19 @@
+"use strict";
+
+module.exports =
+{
+  title: "name",
+  required: [
+    "first",
+    "last"
+  ],
+  type: "object",
+  properties: {
+    last: {
+      type: "string"
+    },
+    first: {
+      type: "string"
+    }
+  },
+};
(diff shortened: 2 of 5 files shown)


jonaslagoni:
So, I am not quite sure of the best cause of action to solve this issue, so might need some pointers from someone who knows the code more in-depth 😄 

Trying to dereference the input in the tests the following happens:
1. First pass, the reference is being dereferenced - https://github.com/APIDevTools/json-schema-ref-parser/blob/f6886ab9ee576754791f2a194b140a637dd98a87/lib/dereference.js#L50
2. Then the pointer is resolved, it says it is a circular reference (and I cant figure out whether that is incorrect behavior) and the resolved value is the root schema when I believe it should be the resolved definition schema `name`.  - https://github.com/APIDevTools/json-schema-ref-parser/blob/f6886ab9ee576754791f2a194b140a637dd98a87/lib/dereference.js#L133
```
pointer = {
  ...,
  path: "/Users/lagoni/Documents/AsyncAPI/json-schema-ref-parser/test/specs/root-internal/root-internal.yaml#/definitions/name",
  originalPath: "/Users/lagoni/Documents/AsyncAPI/json-schema-ref-parser/test/specs/root-internal/root-internal.yaml#/definitions/name",
  value: {
    definitions: {
      name: {
        title: "name",
        type: "object",
        required: [
          "first",
          "last",
        ],
        properties: {
          first: {
            type: "string",
          },
          last: {
            $ref: "#/definitions/name/properties/first",
          },
        },
      },
    },
    $ref: "#/definitions/name",
  },
  circular: true,
  indirections: 0,
}
```
4. Since the pointer is marked as circular, this means that the rest of the references are not crawled, which then means the `$ref` within the definition is not iterated, which to me indicate that there are multiple places that need to change, but I might be wrong 😅 - https://github.com/APIDevTools/json-schema-ref-parser/blob/f6886ab9ee576754791f2a194b140a637dd98a87/lib/dereference.js#L151 

But from this point on, I do not know where or what to try and poke at...

Any suggestions?

cc @philsturgeon (as you are the most recently pushed changed, feel free to re-tag someone else 😄)

jonaslagoni:
With no interest, closing.

PR #248, opened 2021-12-09
Fix tests (by fixing error paths generally) when run in paths with spaces

pimterry:
Previously, on Windows and Linux (comment here was incorrect) these tests failed if the path contained spaces, because the err.source property used %20 instead of a literal space. This happens because internal URL-encoded path representations are attached directly to the error, instead of the actual path.

While this was commented out as a test error, this is actually a real bug - errors referencing paths all contained unexpectedly encoded strings.

(Pro-tip: store all your code in a local folder with spaces, and it's super easy to catch these issues before they become actual bugs that seriously break things. Easy to do in CI too!)

This change ensures that the path representations are at least decoded correctly before any errors are thrown, so that clear path strings are returned.

It would be possible to go further and use `url.toFileSystemPath` to fully transform the URL-formatted path into a standard platform-specific filesystem path, but that's a bigger change, we could be a breaking change for some use cases, e.g. because it would change from unix to Windows paths on Windows.

### DIFF
--- a/lib/pointer.js
+++ b/lib/pointer.js
@@ -91,7 +91,7 @@ Pointer.prototype.resolve = function (obj, options, pathFromRoot) {
     let token = tokens[i];
     if (this.value[token] === undefined || this.value[token] === null) {
       this.value = null;
-      throw new MissingPointerError(token, this.originalPath);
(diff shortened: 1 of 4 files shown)


philsturgeon (approved):
Amazing, thanks Tim!

PR #249, opened 2021-12-09
Allow excluding certain paths from dereferencing

pimterry:
This is the first step towards fixing https://github.com/APIDevTools/swagger-parser/issues/160 and thereby https://github.com/github/rest-api-description/issues/188.

In OpenAPI, examples must be interpreted _literally_. For example, a response could include this `examples` field to illustrate a possible HTTP JSON response body:

```json
"examples": {
  "scim-enterprise-group-list": {
    "value": {
      "schemas": [
        "urn:ietf:params:scim:api:messages:2.0:ListResponse"
      ],
      "totalResults": 2,
      "itemsPerPage": 2,
      "startIndex": 1,
      "Resources": [
        {
          "schemas": [
            "urn:ietf:params:scim:schemas:core:2.0:Group"
          ],
          "id": "abcd27f8-a9aa-11ea-8221-f59b2be9cccc",
          "displayName": "octo-org",
          "members": [
            {
              "value": "92b58aaa-a1d6-11ea-8227-b9ce9e023ccc",
              // v Note the $ref here
              "$ref": "https://api.github.com/scim/v2/enterprises/octo-corp/Users/92b58aaa-a1d6-11ea-8227-b9ce9e023ccc",
              "display": "octocat@github.com"
            }
          ]
        }
      ]
    }
  }
}
```

(This is a smaller version of a real example from [GitHub's OpenAPI spec](https://github.com/github/rest-api-description/))

In this case, this is showing an example where a response really will contain a $ref field - it's not a ref to an external resource in the spec, and it shouldn't ever be dereferenced.

The OpenAPI spec on example objects [here](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#exampleObject) describes this as an "Embedded literal example" and [this discussion](https://github.com/OAI/OpenAPI-Specification/issues/1986#issuecomment-519988505) goes into more detail.

The only way to handle cases like this is to allow callers of `dereference` to stop dereferencing at certain points within a schema's structure. This PR adds a `excludedPathsMatcher` dereference option, which allows the caller to provide a callback that can do this.

The example test shows a quick & dirty approach to how the OpenAPI problem could be solved with this new option. It's not rigorous (really it should be checking against a large list of possible paths from the root) but it shows you the general idea and produces the correct result given the example included here.

### DIFF
--- a/lib/dereference.js
+++ b/lib/dereference.js
@@ -41,8 +41,10 @@ function crawl (obj, path, pathFromRoot, parents, processedObjects, dereferenced
     circular: false
   };
 
+  let isExcludedPath = options.dereference.excludedPathMatcher;
+
   if (options.dereference.circular === "ignore" || !processedObjects.has(obj)) {
-    if (obj && typeof obj === "object" && !ArrayBuffer.isView(obj)) {
+    if (obj && typeof obj === "object" && !ArrayBuffer.isView(obj) && !isExcludedPath(pathFromRoot)) {
       parents.add(obj);
       processedObjects.add(obj);
 
@@ -55,6 +57,9 @@ function crawl (obj, path, pathFromRoot, parents, processedObjects, dereferenced
         for (const key of Object.keys(obj)) {
           let keyPath = Pointer.join(path, key);
           let keyPathFromRoot = Pointer.join(pathFromRoot, key);
+
+          if (isExcludedPath(keyPathFromRoot)) continue;
+
           let value = obj[key];
           let circular = false;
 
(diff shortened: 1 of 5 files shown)


philsturgeon:
@pimterry hey Tim, this looks fantastic. Can you do a quick example in docs/options.md so folks know how to use this feature?

pimterry:
> Can you do a quick example in docs/options.md so folks know how to use this feature?

Done :+1:

alfetopito:
Hi there!
I'm quite interested in this feature.
Any chance to have it released?

Issue #214, opened 2021-02-04
Ability to skip dereferencing inline/internal $ref values


denver-HJS:
Hi all,

[@stoplight/json-ref-resolver](https://github.com/stoplightio/json-ref-resolver)  offers [options](https://github.com/stoplightio/json-ref-resolver/blob/master/src/types.ts#L62-L70) to skip dereferencing either internal or remote references. I've found this feature to be very useful, but now that project is now deprecated and points to this one.

I was wondering if there's already a way to accomplish the same thing using this library?

Essentially I'm hoping to preserve this behavior of taking an example schema:
```json
{
   "paths": {
     "/awesome/endpoint": {
       "post": {
          "responses": {
            "400": {
              "$ref": "#/components/responses/400"
            }
          }
       }
     }
   },
   "components": {
     "responses": {
       "400": {
         "$ref": "https://example.com/path/to/my/oas/file.oas3.json#/components/responses/400"
     }
  }
}
```

and only dereference the remote HTTP reference:

```json
{
   "paths": {
     "/awesome/endpoint": {
       "post": {
          "responses": {
            "400": {
              "$ref": "#/components/responses/400"
            }
          }
       }
     }
   },
   "components": {
     "responses": {
       "400": {
         "description": "The request syntax was malformed and could not be processed.",
          "content": {
             "application/json": {
                "type": "object",
                "properties": {
                  "statusCode": {
                    "type": "integer"
                  },
                  "message": {
                    "type": "string"
                  }
                }
             }
          }
       }
   }
}
```

@P0lip sorry to ping you directly, but I noticed that you've been active in both projects. If you have any pointers it would be much appreciated.

Hopefully this summary makes sense. Thanks!

denver-HJS:
@philsturgeon sorry to mention you directly too, but I see in [one of the issues in the now deprecated Stoplight project](https://github.com/stoplightio/json-ref-resolver/issues/91#issuecomment-692653302) referenced above that you commented that you've been able to meet your requirements with this library. I'm wondering if you have any suggestions for using it to only dereference the remote references and leave the inline ones in place?

philsturgeon:
Why not use bundle if you're ok with references but want them to exist as local refs instead? That's exactly what it's for. :)

denver-HJS:
Hmmm, I thought I tried this but I might have been dereferencing & bundling in some incorrect combination. I'll give that another look. Thanks!

agavazov:
[...]

Issue #145, opened 2020-01-17
"$id" and "$ref" changes in JSON Schema draft 2019-09 and OAS 3.1


handrews:
JSON Schema [draft 2019-09](https://json-schema.org/draft/2019-09/json-schema-core.html) (formerly known as [draft-08](https://json-schema.org/specification-links.html#table-of-all-versions-of-everything)) introduced [several changes](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.8.2) around `$ref` and `$id`.  This draft is [being adopted](https://github.com/OAI/OpenAPI-Specification/pull/1977) by the OpenAPI Specification version 3.1.

* The plain-name fragment declaration function of `$id` was split into a separate keyword, `$anchor`
    * `"$id": "#foo"` becomes `"$anchor": "foo"`
    * note lack of `#t` as the value of `$anchor` is just the name
    * It is still referenced as `"$ref": "#foo"` or `"$ref": "https://example.com/some-schema#foo"`
* `$id` itself MUST resolve to an absolute URI (no fragment)
    * Any schema with an `$id` is therefore a full resource with its own URI, even if embedded in another resource
    * An empty JSON Pointer fragment, e.g. `https://example.com/some-schema#`, is allowed because it is equivalent to not having a fragment
    * This is also because of historical usage in older meta-schemas
* `$ref` can now have other keywords beside it
    * The result of the `$ref` keyword is simply the result of the reference schema
    * That result is combined with other keyword results in the usual way
    * This means that ***it is not any sort of merge***
    * `{"type": "object", "$ref": "foo"}` is equivalent to `{"type": "object", "allOf": [{... contents of foo schema ...}]}`

Since this library has not previously supported `"$id": "#foo"`, if you only support one form I would suggest `"$anchor": "foo"` since that is where OpenAPI is headed.

See also https://github.com/OAI/OpenAPI-Specification/issues/2092 for further details on how this fits into OAS 3.1, as obviously OAS Reference Objects outside of Schema Objects don't know about `allOf`, etc.

Feel free to [ask questions on the JSON Schema slack](https://join.slack.com/t/json-schema/shared_invite/enQtNjc5NTk0MzkzODg5LTVlZGIxNmVhMGY2MWFlYTdiNDQ5NWFiZGUwOThhNmYxZDE0YzA5YjRiOTA5MGY4ZTZlZGZhZDFmYTY4NWM2N2Y).  I will also keep an eye on this issue.

[...]

PR #250, opened 2021-12-09
Ability to skip dereferencing inline/internal $ref values #214

marc0l92:
Add new option to skip the internal references while parsing the document

This pull request fix the issue: https://github.com/APIDevTools/json-schema-ref-parser/issues/214

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -26,7 +26,8 @@ $RefParser.dereference("my-schema.yaml", {
     http: {
       timeout: 2000,                // 2 second timeout
       withCredentials: true,        // Include auth credentials when resolving HTTP references
-    }
+    },
+    skipInternal: false             // Skip internal/local references while parsing the document
   },
   dereference: {
     circular: false                 // Don't allow circular $refs
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -40,7 +40,7 @@ function crawl (parent, key, path, pathFromRoot, indirections, inventory, $refs,
   let obj = key === null ? parent : parent[key];
 
   if (obj && typeof obj === "object" && !ArrayBuffer.isView(obj)) {
-    if ($Ref.isAllowed$Ref(obj)) {
+    if ($Ref.isAllowed$Ref(obj, options)) {
       inventory$Ref(parent, key, path, pathFromRoot, indirections, inventory, $refs, options);
     }
     else {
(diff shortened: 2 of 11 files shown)


philsturgeon:
I'm struggling to see the use case here. So far in #214 it looked like one persons problems were solved with using bundling, then another person was simply using $ref incorrectly (entirely understandable, we've all done it).

Perhaps this is worth merging just for the sake of feature parity with the stoplight tool? 

If so, can we document better? 

What does skip internal mean? If it means "only replace $ref for external files but keep $ref for internal files" that kinda sounds like bundling? Maybe I've been away from my computer for too long or something. Please help me understand what we're trying to do.

marc0l92:
Hello @philsturgeon,
thanks for reviewing this pull request.

You are right to be confused about this issue because it makes the bundling working in a different way, but let me explain you a bit of background.

In my company we created several OpenAPI files using the references in a different way than the official one and we were using a different library bundle them.
The previous library is not supported anymore, so I decided to modify our project to this library to bundle the files because it looks well done and maintained.

The references that this library expect are structured like this:
image

The reference that I'm using in my company are like this:
image

Both approaches have advantages and disadvantages.
For example, the second approach allows you to create links only one layer deep and centralize the versions in a single file (the root). In this way, if a new version of a shared object is released, you need to change it in only one place.
In this example, if you want to migrate to `obj3.v4` you need to change it only once.

Another advantage is that you can share parts of an object between apis. In this example, if two apis wants to share the same obj1 but a different implementation of obj3 they can do it without duplicating the file obj1.

This can be useful is some situations but annoying in others. If you look at `obj1.v1.json` alone you are not able to understand its content until you put it inside a complete api.

Anyway, my goal is not to convince you that this alternative method of linking is a valid one, but explain why the option in this pull request is needed.

My team decided to use the second linking approach long time ago and, if possible, I would like to keep it in this way using this library to do the bundling.

How do you suggest me to document this option in github pages? do you want to rename the option value?

Thanks,
Marco

philsturgeon:
Have you seen this? https://apisyouwonthate.com/blog/json-schema-bundling-finally-formalised/

With $id and new tooling that supports $id in $ref, there is no reason for anyone to work on custom solutions or awkward hacks anymore. We could all just use $id. 

This codebase has some plans to support the new hotness in https://github.com/APIDevTools/json-schema-ref-parser/issues/145, which really is a rewrite in the form of https://github.com/APIDevTools/json-schema-reader, but that rewrite has been paused due to life kicking everyone in the face. Experimentation and PRs welcome.

marc0l92:
Link updated:
https://json-schema.org/blog/posts/bundling-json-schema-compound-documents

PR #266, opened 2022-06-07
feat: a mortdern way to detect browser

Lantianyou:
https://stackoverflow.com/questions/49411796/how-do-i-detect-whether-i-am-on-server-on-client-in-next-js


Lantianyou:
image


Using process in browser will cause app crash, having trouble with it.

philsturgeon:
Thank you for this, this is being implemented as part of larger changes in #280.

PR #268, opened 2022-06-13
Esmification

aerialist7:
Hi Team,
I've prepare the ESM version of the library.
Can you please review the changes and merge them if OK?

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -4,6 +4,8 @@
 
 root: true
 extends: "@jsdevtools"
+parserOptions:
+  sourceType: module
 env:
   node: true
   browser: true
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -2,8 +2,6 @@
 // https://karma-runner.github.io/0.12/config/configuration-file.html
 // https://jstools.dev/karma-config/
 
-"use strict";
-
 const { karmaConfig } = require("@jsdevtools/karma-config");
 const { host } = require("@jsdevtools/host-environment");
 
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -1,10 +1,8 @@
-"use strict";
+import $Ref from "./ref";
+import Pointer from "./pointer";
+import url from "./util/url";
 
-const $Ref = require("./ref");
-const Pointer = require("./pointer");
-const url = require("./util/url");
-
-module.exports = bundle;
+export default bundle;
 
 /**
  * Bundles all external JSON references into the main JSON schema, thus resulting in a schema that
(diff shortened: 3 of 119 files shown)


aerialist7:
Hi @P0lip, hi @philsturgeon

I've fixed the ESLint, but seems tests will fail because of `type: "module"`.

Can you consult me, how to correctly update project config to enable es-modules for test environment?

philsturgeon:
@aerialist7 I understand this situation a lot better now, and I see that there's a way to go ahead with this pull request if you're still interested! 

https://dev.to/bcoe/esm-doesn-t-need-to-break-the-ecosystem-4p8b

They talk about conditional imports: 

```
  {
    "exports": {
      ".": {
        "import": "./index.mjs",
        "require": "./index.cjs"
      },
      "./helpers": {
        "import": "./helpers/helpers.mjs",
        "require": "./helpers/index.js"
      }
    }
  }
```

Can we set that up so things will continue to work for CJS and ESM?

Then we'll avoid the problems discussed in the node-fetch issue.

aerialist7:
Hi @philsturgeon, thank you for your respond. 
I think we can, but we need to configure build to CJS and ESM together.

philsturgeon:
I'll admit I'm not caught up with all this, but I have blundered my way through it with help on other projects. 

https://github.com/stoplightio/spectral-owasp-ruleset/blob/main/package.json

We got that repo building from TS to CJS and ESM so perhaps something can be nicked from there. 

Do you think you could take a swing at it?

philsturgeon:
I've just merged two large changes which will have caused some conflicts here, sorry about that. I'm done merging big changes now though so you've got a clear runway to land this change.

deepakthankachan:
any updates on this? cheers

philsturgeon:
@deepakthankachan you know if this PR has updates because it will show them. Currently there are conflicts and failing tests so we're just waiting for that to get fixed.

wparad:
The best way to proceed with this is first add in building the library with a tool that can output both esm and cjs and then do the esm changes if we want. There shouldn't really be a reason we need to introduce esm though, since esm can already import cjs files which the library currently is. If we see the value in converting to esm in the long run then making those changes piecemeal in follow up PRs is the best way to go.

philsturgeon:
@acunniffe any chance you can pick this PR up and get it over the line?

philsturgeon:
I've made progress over here if anyone is interested in continuing this effort.

philsturgeon:
More progress is being made in #288 and this one isn't being worked on so I'll close this for that.

PR #291, opened 2022-11-26
Chores

philsturgeon:


package.json
```diff
@@ -57,7 +62,6 @@
     "upgrade": "npm-check -u && npm audit fix"
   },
   "devDependencies": {
-    "@amanda-mitchell/semantic-release-npm-multiple": "^2.5.0",
```
philsturgeon, line 60: the unscoped package has been broken for a while so I'm asking @JamesMessinger to mark it deprecated. Let's just do the one thing.

PR #294, opened 2023-01-09
Fix CI tests

jamesmoschou:
Fixes CI tests, as much as possible from a forked repo.

Note: This PR bumps the minimum Node version from 12 to 17. This is because browser tests use an older version of Webpack that requires `NODE_OPTIONS=--openssl-legacy-provider` to be specified when using recent versions of Node. However this option does not exist in older version of Node.


philsturgeon (approved):

philsturgeon:
thanks for this! nearly there, do you know whats up with ubuntu browser tests?

jamesmoschou:
It's failing because it can't launch Edge.

```
09 01 2023 11:25:29.297:INFO [launcher]: Starting browser Edge
09 01 2023 11:25:29.298:ERROR [launcher]: No binary for Edge browser on your platform.
  Please, set "EDGE_BIN" env variable.
```

To be honest I was hoping there would be something different about the Sauce Labs configuration in this repo than my fork that would make this work. Will need to investigate further.

But maybe we don't need it for Edge anyway as it's included in the runner image?
https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md

philsturgeon:
Anything we can do to get away from SauceLabs is a good thing IMO. It's been a nightmare trying to keep it happy.

jamesmoschou:
Ok that's good to know.

These commits should now pass. https://github.com/criteria-labs/json-schema-ref-parser/actions/runs/3888908655

I got rid of Sauce Labs and browsers are now just tested on their "home" platform, i.e. Edge on windows, Safari on macOS. Though there is no runner for macos-latest because the karma launcher for Safari doesn't work (it hasn't been updated in 9 years). So there are no Safari tests at all.

philsturgeon (approved):
Looming good. No bother about safari, it was already disabled.

zomars:
This is breaking most apps that rely on Node 16:

image

Elhebert:
We also had our app breaking without warning due to this.

I feel like this should have been released as a new major version. Changing the minimum Node version is a breaking change after all.

package.json
```diff
@@ -43,7 +43,7 @@
     "fs": false
   },
   "engines": {
-    "node": ">= 14"
+    "node": ">= 17"
```
AviVahl, line 46: @jamesmoschou this caused a regression on our end as well. just having it in the tree of a node v16.x project breaks during dependency installation (yarn, in our case), even if we don't directly use this package.

PR #295, opened 2023-01-09
ES Modules

jamesmoschou:
(This PR builds off https://github.com/APIDevTools/json-schema-ref-parser/pull/294)

Convert the project to ES Modules and implements dual CommonJS/ES module exports.


jamesmoschou:
Hey @philsturgeon this now implements dual CommonJS/ES module support and is ready to merge.

The one thing to call out is that the minimum Node version is now 17, not 14 according to the engines property in package.json. Though it only matters for running tests and it should still work in 14.

philsturgeon:
I’m ok increasing versions under those terms as we’re aiming for EAM in a major anyway. Now we’ve merged the other PR can we update and see if this is good to go? 

Thank you so much for your support.

jamesmoschou:
Yep, looks like all checks have passed

jamesmoschou:
oh there's conflicts...

jamesmoschou:
ok fixed

philsturgeon (approved):

philsturgeon:
Thank you @jamesmoschou! This is a great update. 

I had to manually release it because semantic-release has gone off the rails, but hopefully thats working ok.

Would you be up for maintaining this package because I am working in the woods 90% of the time now and don't want to be stressing about this when I get back to a computer. 

I'm also asking if https://github.com/readmeio/json-schema-ref-parser/issues/64 can take over, which may just involve merging some of the recent changes and me pointing this package to them.

jamesmoschou:
Hey @philsturgeon I can help maintain this package. Though I mean maintain in the literal sense, I wouldn't be developing any new feature requests.

Looks like the node engine breaking change snuck in to v9.* version instead of v10. I can update this is you set up the permission :)

coffeebe4code:
Sorry, referencing my comment here.
https://github.com/APIDevTools/json-schema-ref-parser/pull/300#issuecomment-1400809753

It looks like this is the one that broke our project.

philsturgeon:
@coffeebe4code fixed in 9.1.2.

PR #299, opened 2023-01-20
fix: change type of onDereference to non-required

jonluca:
The type of onDereference is required, which breaks types in some downstream packages. The code also checks for its existence on L72 of dereference.js so this is fine.


jonluca:
Also @philsturgeon I'm happy to takeover this project. I was meaning to rewrite in typescript one of these days anyways

bcherny:
This change looks good to me!

philsturgeon:
@jonluca thank you for the fix and the offer! I’ve added you as a maintainer as well as @jamesmoschou, so the two of you could share responsibilities and review each other’s work. I’ll be around if you need anything, same username on most of the socials.

philsturgeon (approved):

Issue #285, opened 2022-10-26
Need more maintainers / contributors


philsturgeon:
Hey folks, trying to get a v10 out at the moment but we're blocked on #268 and #283 so if you can get to that I'd really appreciate it. 

As a larger thing, I am struggling with doing far too many jobs, including trying to grow an entire reforestation charity which is planting thousands of trees and managing multiple woodlands all over the country. It's a LOT and I cannot be throwing any/much more time on maintaining OSS stuff I don't even use for anything.

See more over here: https://phil.tech/2022/bundling-openapi-with-javascript/

tldr we need to either 

1. get some enthusiastic fresh blood in here as maintainers.
2. find an alternative library that's got feature party and I'll start the deprecation and redirection process, helping write migration guides.
3. revive the [json-schema-reader](https://github.com/APIDevTools/json-schema-reader) rewrite to handle big ticket items like #145.
4. y'all start throwing a whole load of [sponsorship money in here](https://github.com/sponsors/philsturgeon) and I'll contract out the work.

Discuss!

braekevelt:
@philsturgeon Any chance you would consider releasing commit cd9bd91b64587916f13e4f3bcf1e9024777d6697 as v9.0.10 while you are working on v10? According to the CI checks of #247 it is quite stable at that point, at least for node environments. Browsers could still stay at v9.0.9 if they want to, like it is now.

philsturgeon:
There is a new effort from the JSON Schema crew on getting everyone aligned on what $ref is meant to be about, how it should work across OAS and JSON Schema, etc. 

https://github.com/json-schema-org/referencing

Probably relevnt. No brainspace to try.

philsturgeon:
Specifically first I would appreciate help getting #288 over the line, more context #290.

jonaslagoni:
We are still discussing what to do for AsyncAPI (regarding references in general) but wanted to give you our current dilemma, before your deadline, to get your thoughts on the situation @philsturgeon.

The main reason we are looking towards this library is that we need it for our [parser](https://github.com/asyncapi/parser-js/) (indirectly in Spectral, which we also help maintain). 

[You can find the full discussion we are having here, regarding requirements for ref resolvers in general, both from the API and general functional and non-functional requirements](https://github.com/asyncapi/community/discussions/485), but here is the TLDR (if we were to help out, as the library in its current state dont quite fit into our use-case):

[...]

PR #300, opened 2023-01-22
Breaking Change: Rewrite repo in typescript, change tests to vitest, fix small bugs

jonluca:
I rewrote the repo in typescript and changed the testing scheme to use vitest. I also bumped some dependencies and removed others.

This is a pretty major refactor but all tests still pass. 

I also undid the `type: module` and made the required node version the stable one (node 16) instead of 17. 

Important differences:

* Vitest now uses the builtin c8 coverage tool
* The special characters test does not currently work due to a bug in vites filesystem parsing, seen here https://github.com/vitest-dev/vitest/issues/2731 - this shouldn't impact the actual logic of the app, and will work outside of vite
* I've changed browser compatibility to use jsdom environment in node instead of using karma. This gets us a _little_ less reliability in exchange for a lot more speed in testing. Also, with modern browser envs, I feel like needing the entirety of karma is overkill
* I've added prettier and some more eslint rules. 
* I've changed the default package manager to yarn


wparad:
I feel like this is going to be impossible to review. How do you expect the review to happen with:
image

jonluca:
I've done this with @philsturgeon for a few repos in the OpenAPI space. I'm also comfortable taking over as maintainer given https://github.com/APIDevTools/json-schema-ref-parser/issues/285

package.json
```diff
@@ -37,67 +41,68 @@
   },
   "license": "MIT",
   "funding": "https://github.com/sponsors/philsturgeon",
-  "main": "lib/index.js",
-  "typings": "lib/index.d.ts",
-  "exports": {
```
philsturgeon, line 42: Will this provide the same ES modules + CJS support that was added in #295? I don't want to annoy older users, but perhaps if we've raised the node version high enough then thats moot now anyway?

package.json
```diff
@@ -37,67 +41,68 @@
...
+    "test:update": "vitest -u",
+    "test:watch": "vitest -w",
     "upgrade": "npm-check -u && npm audit fix"
   },
   "devDependencies": {
-    "@babel/polyfill": "^7.12.1",
-    "@chiragrupani/karma-chromium-edge-launcher": "^2.2.2",
-    "@jsdevtools/eslint-config": "^1.0.7",
     "@jsdevtools/host-environment": "^2.1.2",
-    "@jsdevtools/karma-config": "^3.1.7",
```
philsturgeon, line 76: Thank you, this is helping me deprecate those older @jsdevtools packages that were mainly used for APIDevTools packages. https://www.npmjs.com/package/@jsdevtools/karma-config

philsturgeon:
From a user perspective is this is breaking change or is this a huge tech debt cleanup? We just released v10 so if its not a BC for the user then v10.1 would be good, but a v11 would be welcome if this is legitimately a breaking change.

Also semantic release is all screwy so please make sure thats working before v11 or v12 is out.

package.json
jonluca: I can remove the host-env one too, since the environment is always running in node.

package.json
jonluca: Yes, I moved this back to being a cjs package. I also changed the pinned version from node experimental (17) to LTS (16). 14 will EOL in April so we should still support 16.

jonluca:
@philsturgeon I think this should be a breaking change. While all tests pass this is a significant enough refactor I'd feel most comfortable with a major version bump.

It also changes things like the required engine and dependency versions, and there might be weirdness in some peoples systems. I've tested it as thoroughly as I can think, but it doesn't feel like a 0.x bump

coffeebe4code:
Am I understanding that this bumped the repo to 9.1.1 as well?

We have a breaking change in our code. We have had to go back to 9.0.9.
9.1.0 would probably be fine, but we no longer wanted to receive updates with ~ so we are going back to 9.0.9.

Has this project moved to new maintainers?

philsturgeon:
@coffeebe4code some changes from main snuck into 9.1.1 when I had to manually tag it due to semantic release going wonkier than I've ever known it to go. 9.1.2 solved the problem. That's not related to v10 or this PR.

@jonluca fair enough, go for it. v11!

@jamesmoschou tagging you in on this one. 

/fades away into the nearest bush.

jonluca:
@jamesmoschou can you please take a look? Want to make sure this works + gets out before I work on the next set of PRs

jamesmoschou:
@jonluca what's the benefit of moving to yarn?

I'm worried it will introduce another dependency for people. Whereas npm comes with node out of the box and I think has gotten a lot better since yarn originally came about.

jonluca:
No real reason I suppose, I havent revisited npm since moving to yarn a few years ago and its my default for new projects. Happy to move back to npm if that's an issue though

lib/index.ts
```diff
@@ -0,0 +1,413 @@
...
+   * It just reads a single file in JSON or YAML format, and parse it as a JavaScript object.
+   *
+   * @param [path] - The file path or URL of the JSON schema
+   * @param [schema] - A JSON schema object. This object will be used instead of reading from `path`.
+   * @param [options] - Options that determine how the schema is parsed
+   * @param [callback] - An error-first callback. The second parameter is the parsed JSON schema object.
+   * @returns - The returned promise resolves with the parsed JSON schema object.
+   */
+  public parse(schema: RefParserSchema): Promise<JSONSchema>;
+  public parse(schema: RefParserSchema, callback: SchemaCallback): Promise<void>;
```
jamesmoschou, line 69: I've not done method overloading in TypeScript before. Do they all have to have the same return value, or should the variants with a callback return void instead of Promise<void>?

jonluca: Well the types that are currently used in the repo are actually incorrect - all the function calls return a Promise, since it's an async function. This just types it correctly.

However, with this type-definition overloading, it's saying that when you pass in just `schema`, the promise will resolve to a `JSONSchema`, whereas if you pass in a callback the callback will be called and the promise will be void.

I don't really like the usage of `arguments` in all these functions, and it would be nice to be able to write new APIs that actually do satisfy some of these apis. They're written in a very "accepting" way right now, IMO.

lib/refs.ts
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  /**
+   * Resolves the given JSON reference and returns the resolved value.
+   *
+   * @param path - The path being resolved, with a JSON pointer in the hash
+   * @param [options]
+   * @returns - Returns the resolved value
+   */
+  get(path: string, options?: $RefParserOptions): JSONSchema4Type | JSONSchema6Type | JSONSchema7Type {
```
jamesmoschou, line 89: The return value here might not be JSON Schema. If this library used used on an OpenAPI document for example, it could be any one of these https://spec.openapis.org/oas/v3.1.0#components-object.

jamesmoschou:
I only went through the lib files and had a few questions. I'm assuming the changes to .spec.ts files are just ts/prettier changes?

Reading up a bit more, I think yarn is fine as it just affects contributors, `npm install` should still work.

Soo happy to get rid of Karma, it was an absolute pain.

jamesmoschou:
Also I'm wondering if we should sort out https://github.com/APIDevTools/json-schema-ref-parser/issues/303 first?

lib/ref.ts
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  /**
+   * Determines whether the given value is a JSON reference.
+   *
+   * @param value - The value to inspect
+   * @returns
+   */
+  static is$Ref(value: any): value is { $ref: string; length?: number } {
+    return value && typeof value === "object" && typeof value.$ref === "string";
```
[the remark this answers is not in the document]
jonluca: I dont _think_ so but I'll add it back

jonluca:
Yes the spec changes are:

* Prettier
* Slight changes due to vitest differences
* Typescript additions

Contents of the tests are identical, with one added test

And regarding #303 - I think this was an issue with moving to an esm + cjs build. This rewrite should work out of the box

capture 2023-01-30 at 10 00 25 AM

This is an example of that usecase where it works.

lib/refs.ts
jamesmoschou: ahh you're right, sorry.

jamesmoschou:
Thanks for clarifying my questions, happy to merge.

Do we want to next version to be 10.1 or 11.0? I think you and @philsturgeon were discussing a major version bump? If so, I think semantic-release is going to deploy 10.1 as-is without a `BREAKING CHANGE:` commit.

jonluca:
If we squash the commit won't it use the title of the PR? I could be mistaken, I'll push an update tomorrow morning if not

jamesmoschou:
Ahh, squashing will work. But it's more of a GitHub feature to use the PR name than the semantic-release feature
https://semantic-release.gitbook.io/semantic-release/support/troubleshooting#squashed-commits-are-ignored-by-semantic-release

approving!

jamesmoschou (approved):

philsturgeon:
Well shit, it was v10.1 anyway. LMK if I gotta nuke it from orbit.

jonluca:
Ha I was wrong, no it's ok, we can keep it as is. I'll be making a few more PRs in the coming days so we can just move forward then

PR #315, opened 2023-03-28
Fix handling of Windows absolute paths. 

### LINKED ISSUE #311: Incorrect path resolving for absolut paths under windows
Hello everyone,

unfortunately I noticed that the resolving of absolute paths under windows is not working for me. The code doesn't seem to recognize that a absolute path is entered and prepends another absolute path. 

I tracked the error down to this function (https://github.com/APIDevTools/json-schema-ref-parser/blob/main/lib/util/url.ts#L177). Both checks `hasProjectDir` and `hasProjectUri` evaluate to false as my api.yaml isn't part of this project. `projectDir` evaluates to `C:\Projekte\projectname\node_modules\@apidevtools\json-schema-ref-parser\dist\` in my setup but only `C:\Projekte\projectname` is shared between the library and the source code of the project. 

It would be great if the problem could be fixed!

Here is also the stacktrace where you can see, what happend to the absolute path.
```json
{
  "stack":"JSONParserError: Error opening file "C:\Projekte\projectname\node_modules\@apidevtools\json-schema-ref-parser\dist\C:\Projekte\projectname\dist\server\api.yaml" ENOENT: no such file or directory, open 'C:\Projekte\projectname\node_modules\@apidevtools\json-schema-ref-parser\dist\C:\Projekte\projectname\dist\server\api.yaml' at Object.<anonymous> (C:\Projekte\projectname\node_modules\@apidevtools\json-schema-ref-parser\dist\lib\resolvers\file.js:72:23) at Generator.throw (<anonymous>) at rejected (C:\Projekte\projectname\node_modules\@apidevtools\json-schema-ref-parser\dist\lib\resolvers\file.js:29:65)",
  "code":"ERESOLVER",
  "name":"ResolverError",
  "message":"Error opening file "C:\Projekte\projectname\node_modules\@apidevtools\json-schema-ref-parser\dist\C:\Projekte\projectname\dist\server\api.yaml" ENOENT: no such file or directory, open 'C:\Projekte\projectname\node_modules\@apidevtools\json-schema-ref-parser\dist\C:\Projekte\projectname\dist\server\api.yaml'",
  "source":"C:\Projekte\projectname\node_modules\@apidevtools\json-schema-ref-parser\dist\C:\Projekte\projectname\dist\server\api.yaml",
  "path":null,
  "ioErrorCode":"ENOENT",
  "footprint":"null+C:\Projekte\projectname\node_modules\@apidevtools\json-schema-ref-parser\dist\C:\Projekte\projectname\dist\server\api.yaml+ERESOLVER+Error opening file "C:\Projekte\projectname\node_modules\@apidevtools\json-schema-ref-parser\dist\C:\Projekte\projectname\dist\server\api.yaml" ENOENT: no such file or directory, open 'C:\Projekte\projectname\node_modules\@apidevtools\json-schema-ref-parser\dist\C:\Projekte\projectname\dist\server\api.yaml'"
}
```

Used version of apidevtools: `"@apidevtools/json-schema-ref-parser": "10.1.0"`

matt1097:
This should fix #311 
Added handling for absolute Windows file paths.

Added some additional tests around the url.fromFileSystemPath() to help catch this error in the future as no existing test was failing.

### DIFF
--- a/lib/util/is-windows.ts
+++ b/lib/util/is-windows.ts
@@ -1,0 +1,7 @@
+/**
+ * Returns if the system is a Windows system or not
+ * @returns 
+ */
+export function isWindows() {
+    return /^win/.test(globalThis.process ? globalThis.process.platform : "");
+}
--- a/lib/refs.ts
+++ b/lib/refs.ts
@@ -1,12 +1,13 @@
 import { ono } from "@jsdevtools/ono";
 import $Ref from "./ref.js";
 import * as url from "./util/url.js";
+import { isWindows } from "./util/is-windows.js";
 import type { JSONSchema4Type, JSONSchema6Type, JSONSchema7Type } from "json-schema";
 import type { JSONSchema } from "./types/index.js";
 import type $RefParserOptions from "./options.js";
 
-const isWindows = /^win/.test(globalThis.process ? globalThis.process.platform : "");
-const getPathFromOs = (filePath: string): string => (isWindows ? filePath.replace(/\\/g, "/") : filePath);
+
+const getPathFromOs = (filePath: string): string => (isWindows() ? filePath.replace(/\\/g, "/") : filePath);
 
 interface $RefsMap {
   [url: string]: $Ref;
--- a/lib/util/url.ts
+++ b/lib/util/url.ts
@@ -1,9 +1,9 @@
-const isWindows = /^win/.test(globalThis.process ? globalThis.process.platform : ""),
-  forwardSlashPattern = /\//g,
+const forwardSlashPattern = /\//g,
   protocolPattern = /^(\w{2,}):\/\//i,
   jsonPointerSlash = /~1/g,
   jsonPointerTilde = /~0/g;
-import { join } from "path";
+import { join, isAbsolute } from "path";
+import { isWindows } from "./is-windows";
 
 const projectDir = join(__dirname, "..", "..");
 // RegExp patterns to URL-encode special characters in local filesystem paths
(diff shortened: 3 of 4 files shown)


meepen:
Kind of wondering, why is this library doing all this weird stuff with checking for windows and manual path fixing when there's available libraries that do this all for us? For example, node's in-built [path](https://nodejs.org/api/path.html) library

sukeshpabolu:
Hi please merge this if possible

gocarlos:
@jonluca anything preventing this from being merged?

jonluca:
> Kind of wondering, why is this library doing all this weird stuff with checking for windows and manual path fixing when there's available libraries that do this all for us? For example, node's in-built [path](https://nodejs.org/api/path.html) library

Yeah I agree - I think I'd prefer to just go through and fix all these issues directly. Will take a look right now, and if I don't see a fast way of fixing it I'll merge in this PR.

lib/util/is-windows.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Returns if the system is a Windows system or not
+ * @returns 
+ */
+export function isWindows() {
+    return /^win/.test(globalThis.process ? globalThis.process.platform : "");
```
jonluca, line 6: is there a reason this is a function now, and not a constant? globalThis.process should never change during runtime?

jonluca (approved):

lib/util/is-windows.ts
jonluca: ah for the test, I see

jonluca:
Thanks for the PR - I'm going to release a breaking change soon with changes that should fix what this is targeting. In general I think we should revamp the windows logic, like @meepen said.

jonluca:
https://github.com/APIDevTools/json-schema-ref-parser/pull/321

PR #317, opened 2023-06-02
fix: External refs with overrides

simonbothen-radinn:
The external resolver did not take into account for overrides. E.g take the following example:

```
$ref: 'base.yaml'
id: my_id
settings:
  $ref: 'base_settings.yaml'
  useCache: true
```

This failed to resolve on `$ref: 'base_settings.yaml'`. This is due to to root object passes `$Ref.isExternal$Ref(obj)` and therefore not considering the other child objects. We should still crawl the tree even though the root has a $ref.


jonluca:
Included in https://github.com/APIDevTools/json-schema-ref-parser/pull/321

Issue #200, opened 2020-12-11
Fails to resolve correct path/filename for extended $ref


daniellubovich:
I recently ran into an issue testing some of my more complex schema documents where combination schema $refs cannot be resolved properly if they are part of an "extended" $ref. It is a little hard to explain the case, but I think I have found a simple enough schema that causes the problem to occur.

```
// base.json
{
  "type": "object",
  "properties": {
    "some_value": { "$ref": "defs.json#/definitions/astring" }
  }
}

//

{
  "definitions": {
    "astring": {
      "description": "astring",
      "$ref": "defs2.json#/definitions/bstring"
    }
  }
}


// defs2.json
{
  "definitions": {
    "bstring": {
      "oneOf": [
        {
          "$ref": "#/definitions/cstring"
        },
        {
          "$ref": "#/definitions/dstring"
        }
      ]
    },
    "cstring": {
      "type": "string"
    },
    "dstring": {
      "type": "number"
    }
  }
}
```

The error that I see when trying to resolve this looks like:
`Error while processing route: dynamic-form Token "cstring" does not exist. MissingPointerError: Token "cstring" does not exist.`

This only seems to happen when the intermediate schema object is an "extended" ref. For example, if we remove the description from `astring` then we are able to dereference without issue. From doing some debugging it looks to me like our issue is that when resolving `cstring`, the path we resolve is `defs.json#/definitions/cstring` instead of looking in `defs2.json`. I think this is because when we resolve extended $refs, we explicitly do not update the pointer's path.

For reference:
https://github.com/APIDevTools/json-schema-ref-parser/blob/13b00923fb279cfcb2a72e0906f3eadc13199d08/lib/pointer.js#L240-L245

I would like to propose a fix, but feel ill-equipped to do so without some guidance. It does seem like updating the path in the above code fixes my specific issue, but clearly that decision was made for a reason so I'd rather not step on any toes without understanding the intent. Any thoughts would be greatly appreciated!

daniellubovich:
I should also mention the issue described here: https://github.com/APIDevTools/json-schema-ref-parser/issues/168.

Based on the draft 07 spec, it seems like the behavior of objects with $ref and additional properties appears to be undefined. I think the decision on how to proceed with this behavior going forward may inform the decision made for this issue as well.

karlvr:
This feels like the same problem that I've had, but I'm not sure what an extended ref is!

The problem I've had is that when it comes time to resolve (using your example):

```
[...]

PR #321, opened 2023-09-18
Fix windows paths and external refs

jonluca:
Adding in the code to fix some path issues and some relative paths.

This doesn't fully address #200 - we really need a more updated test schema.

I imported a majority of the code from that PR, though, and will take a look at some point in the future. 

https://github.com/APIDevTools/json-schema-ref-parser/pull/315 windows absolute paths should be included here


sukeshpabolu (approved):
ok

PR #341, opened 2024-04-03
chore: inherit encoding type from Plugin for text parsing option

liamc-altruistiq:
When compiled, there is a type error that makes builds fail on a package that I would like to use this as a dependency. It's related to `Plugin.parse.text` which appears to have some overriding typing for the `encoding` option. This just removes that override and instead inherits the `BufferEncoding` type from `Plugin`.


liamc-altruistiq:
@jonluca will a force-release be necessary here? the CI action claims that nothing of note has changed (which I guess is true) but the built `dist` files in the package will still contain the invalid types that are of concern.

jonluca:
Yeah need to make a commit with a `fix: `, I'll take a look

PR #347, opened 2024-04-24
fix: Enable compilerOptions noImplicitReturns and noUnusedParameters to improve type safety

tekener:
Hi, we are using your lib inside a typescript project having the following quite common compilerOptions enabled inside out tsconfig.json.

```
    "noImplicitReturns": true,
    "noUnusedParameters": true
```

This PR enabled this settings on your project as well because the code issues hiding in your code are letting our tsc fail as well. To enable it I fixed most of the underlying code issues. I added `@ts-ignore` to the `PluginResult` interface because I was too shy to remove the not used generic parameter `O`.

Thank you

### DIFF
--- a/lib/bundle.ts
+++ b/lib/bundle.ts
@@ -290,6 +290,7 @@ function findInInventory(inventory: InventoryEntry[], $refParent: any, $refKey: 
       return existingEntry;
     }
   }
+  return undefined
 }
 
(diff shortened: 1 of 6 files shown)


tekener:
Hi @jonluca, just kindly asking if you are creating a release for that.

PR #351, opened 2024-06-03
code optimisation for circular references issue

SB-rohitdesai:
Getting 500 Error or Sometimes might need to wait a long time before getting this error when spec having circular references


jonluca:
This branch is completely divergent from HEAD. If you can isolate the changes into a PR that touches just the typescript files and adds tests I would merge that in.

PR #360, opened 2024-12-05
Improve missing pointer error message

hudlow:
It's much more helpful if the missing pointer error tells you what the pointer is.


jonluca:
This is great, thanks!

Issue #358, opened 2024-09-20
Add function call  when circular reference is found


jopech-tibco:
Beside of ignoring or throwing error, could be really nice if you could executed a call back function when a circular reference is found. A call similar to `onDereference` .

My use is that I need to find all the circular reference from a given schema, currently there is no way to do it

PR #366, opened 2025-01-22
feat: creation of a new `onCircular` hook for accumulating circular refs

erunion:
This creates a new `options.dereference.onCircular` hook that will be executed when a circular `$ref` is detected, and similar to `onDereference`, will allow consumers to accumulate a list of circular references within their schemas.

I have written a test for this, and also updated a few test case typos where they were using the legacy `options.$refs` property instead of `options.dereference`.

Closes out https://github.com/APIDevTools/json-schema-ref-parser/issues/358


jonluca (approved):

jonluca:
Looks great! Thanks for the PR

PR #193, opened 2020-09-10
add more details on missing $ref

AlexHolly:
Example Current infos:
https://apitools.dev/swagger-parser/online/

```
openapi: 3.0.2
info:
  description: >-
    This is a sample server Petstore server.
  version: 1.0.0
  title: Swagger Petstore
  termsOfService: 'http://swagger.io/terms/'
  contact:
    email: apiteam@swagger.io
  license:
    name: Apache 2.0
    url: 'http://www.apache.org/licenses/LICENSE-2.0.html'
paths:
  /pet:
    post:
      tags:
        - pet
      parameters:
        - $ref: '#/components/parameters/ThisIsMissing'
components:
  parameters:
    petId:
      name: petId
      in: path
      required: true
      schema:
        type: integer
        format: int64
```

PR will give this data

```
$ref value of: #/paths/~1pet/post/parameters/0 
  is missing target: #/components/parameters/ThisIsMissing
  found: #/components/parameters
```

PR #368, opened 2025-01-26
Add more details on missing $ref

AlexHolly:
Old PR https://github.com/APIDevTools/json-schema-ref-parser/pull/193


lib/pointer.ts
```diff
@@ -120,10 +122,23 @@ class Pointer<S extends object = JSONSchema, O extends ParserOptions<S> = Parser
         }
 
         this.value = null;
-        throw new MissingPointerError(token, decodeURI(this.originalPath));
+
+        let path: any = '';
+
+        if (path !== undefined) {
```
jonluca, line 128: path will never be undefined since you define it two lines above? Also please dont use `: any`

jonluca:
LGTM, thanks!

Issue #365, opened 2025-01-22
Dereferencing behavior on OpenAPI 3.1 schemas


erunion:
Hi there! We have a fork that we're trying to wind down with the updates that yall have been doing here and one hang up we have is regarding dereferencing OpenAPI 3.1 schemas. Within the specification they allow you to have `summary` and `description` live alongside a `$ref`, with that `summary` and `description` taking precedence over any that may exist within the referenced schema.

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#reference-object

Looking at the new `onDereference` handler it at first glance seems like it would be be able to support this but it neither allows you to mutate the schema you have, as it's called after the schema has been dereferneced[^1], nor does it surface the other schema elements to `onDereference` that live alongside the `$ref` property:

```ts
import $RefParser from ".";

const schema = {
  required: ['name'],
  type: 'object',
  definitions: {
    name: {
      type: 'string',
      description: "Someone's name",
    },
  },
  properties: {
    name: {
      $ref: '#/definitions/name',
    },
    secretName: {
      $ref: '#/definitions/name',
      description: "Someone's secret name",
    },
  },
  title: 'Person',
};

(async () => {
  await $RefParser.dereference(schema, {
    dereference: {
      onDereference: (path: any, value: any, parent: any, parentPropName: any) => {
        console.log('[onDereference]', { path, value, parent, parentPropName });
      },
    }
    });
})();
```

```
[onDereference] {
  path: '#/definitions/name',
  value: { type: 'string', description: "Someone's name" },
  parent: {
    name: { type: 'string', description: "Someone's name" },
    secretName: {
      '$ref': '#/definitions/name',
      description: "Someone's secret name"
    }
  },
  parentPropName: 'name'
}
[onDereference] {
  path: '#/definitions/name',
  value: { type: 'string', description: "Someone's name" },
  parent: {
    name: { type: 'string', description: "Someone's name" },
    secretName: { type: 'string', description: "Someone's name" }
  },
  parentPropName: 'secretName'
}
```

Because this work is exclusive to OpenAPI 3.1 schemas I don't know if it makes a whole lot of sense to be the default dereferencing behavior here, but do you have any ideas on how we could be able to support this within this library? We would love to get rid of our fork (that supports this specific behavior) so if we could maybe pull this in via an opt-in option of some sort, if you're into that, we'd happily PR it.
 

[...]

PR #369, opened 2025-01-28
feat: `dereference.preservedProperties` for preserving data during dereferencing

erunion:
This introduces a new `dereference.preservedProperties` configuration that allows for custom `$ref` dereferencing behavior to happen where if you want to preserve, and prioritize an existing, property you can.

The use case for this is theOpenAPI 3.1 specification where because it allows you to define a `description` and/or `summary` alongside a `$ref` pointer, and require that that `description` and `summary` take precedence over any that may be inside of the referenced schema there is currently no way to handle this within the library currently.

For example, the following schema:

```js
{
  required: ['name'],
  type: 'object',
  definitions: {
    name: {
      type: 'string',
      description: "Someone's name",
    },
  },
  properties: {
    name: {
      $ref: '#/definitions/name',
    },
    secretName: {
      $ref: '#/definitions/name',
      description: "Someone's secret name",
    },
  },
  title: 'Person',
}
```

With the existing dereferencing behavior this schema is dereferenced into the following:

```js
{
  required: [ 'name' ],
  type: 'object',
  definitions: { name: { type: 'string', description: "Someone's name" } },
  properties: {
    name: { type: 'string', description: "Someone's name" },
    secretName: { type: 'string', description: "Someone's name" }
  },
  title: 'Person'
}
```

However, because in an OpenAPI 3.1 definition I would want to use "Someone's secret name" instead of "Someone's name" on the `secretName` property, using this new configuration I can do that:

```ts
await $RefParser.dereference(schema, {
  dereference: {
    preservedProperties: ['description'],
  }
});
```

```js
{
  required: [ 'name' ],
  type: 'object',
  definitions: { name: { type: 'string', description: "Someone's name" } },
  properties: {
    name: { type: 'string', description: "Someone's name" },
    secretName: { type: 'string', description: "Someone's secret name" }
  },
  title: 'Person'
}
```

1. **Why make this configurable instead of the default behavior?** Because this seems to be specific to OpenAPI 3.1 and it could potentially be a big breaking change to how people are currently handling these sorts of cases.
2. **This config is generic and has nothing specific to OpenAPI 3.1 though.** Correct. This behavior would be handled upstream in consumers of the library (like `swagger-parser`) and conditionally feed in this config for OpenAPI 3.1 definitions.

Closes https://github.com/APIDevTools/json-schema-ref-parser/issues/365


lib/dereference.ts
```diff
@@ -123,7 +123,27 @@ function crawl<S extends object = JSONSchema, O extends ParserOptions<S> = Parse
             circular = dereferenced.circular;
             // Avoid pointless mutations; breaks frozen objects to no profit
             if (obj[key] !== dereferenced.value) {
+              // If we have properties we want to preserve from our derefernced schema then we need
```
jonluca, line 126: nit: typo

erunion: ope

lib/dereference.ts
```diff
@@ -123,7 +123,27 @@ function crawl<S extends object = JSONSchema, O extends ParserOptions<S> = Parse
...
+              // to copy them over to our new object.
+              const preserved: Map<string, unknown> = new Map();
```
jonluca, line 128: nit: can we wrap this logic and this object allocation inside a check for if `derefOptions?.preservedProperties` is set? if its not set we can skip a lot of this logic.

jonluca:
This lgtm - if you can address the minor comment then I'll merge it in. Thanks!

lib/dereference.ts
erunion: Absolutely!

erunion:
@jonluca I've just pushed up your recs. Thanks for the speedy review! 💨

PR #373, opened 2025-02-26
fix: `http.timeout` default does not match the documentation

erunion:
Ran into a fun problem today where while running a test case that was processing 4,000+ schemas my tests began to slow to a crawl. Thinking it was a `fetch` problem I dug into the documentation and noticed that the docs mention `http.timeout` having a 5s default.

Turns out it's actually 60s! 😰 

~This drops the `http.timeout` default down to 5s in order to match the docs.~


jonluca:
We should probably change it so that the docs say that the timeout is 60s rather than changing the default. This release might break peoples existing workflows. 

I think a Readme / docs update is better

erunion:
@jonluca Sounds good! Just pushed that up.

jonluca:
awesome, thanks!

PR #374, opened 2025-02-27
fix: allow `$ref` pointers to point to a `null` value

erunion:
I've run into a curious issue where the following schema throws a `MissingPointerError` exception:

```yaml
components:
  examples:
    product:
      value:
        data:
          admission:
            $ref: "#/components/examples/product/value/data/pas"
          pas: null
```

I'm having a hard time understanding _why_ the `Pointer` class is running this substring lookup for if a token _does_ exist but it's `null`, but because the token we're looking for is the last in the list in this block it never ends up running the `for` loop below this line to look for a substring -- resulting in it falling through to throw a `MissingPointerError`.

https://github.com/APIDevTools/json-schema-ref-parser/blob/5303da99a58089ddaa4ae08f01205ae9df102a67/lib/pointer.ts#L108-L122

I've managed to fix this by doing a secondary "does this token exist and is it null? ok return that." check and that has resolved this problem in the dereferencer but I also don't know if this is the best way to fix this, or if it will have downstream effects (no tests break with it), or would be considered as a breaking change.

Changing line 108 to **only** check `this.value[token] === undefined` also fixes this issue but even though no tests break with that I don't know if it's the right fix.

Looking for guidance here!

### DIFF
--- a/lib/pointer.ts
+++ b/lib/pointer.ts
@@ -5,6 +5,8 @@ import * as url from "./util/url.js";
 import { JSONParserError, InvalidPointerError, MissingPointerError, isHandledError } from "./util/errors.js";
 import type { JSONSchema } from "./types";
 
+export const nullSymbol = Symbol('null');
+
 const slashes = /\//g;
 const tildes = /~/g;
 const escapedSlash = /~1/g;
@@ -125,7 +127,9 @@ class Pointer<S extends object = JSONSchema, O extends ParserOptions<S> = Parser
         // actually instead pointing to an existing `null` value then we should use that
         // `null` value.
         if (token in this.value && this.value[token] === null) {
-          this.value = null;
+          // We use a `null` symbol for internal tracking to differntiate between a general `null`
+          // value and our expected `null` value.
+          this.value = nullSymbol;
           continue;
         }
 
--- a/lib/pointer.ts
+++ b/lib/pointer.ts
@@ -121,6 +121,14 @@ class Pointer<S extends object = JSONSchema, O extends ParserOptions<S> = Parser
           continue;
         }
 
+        // If the token we're looking for ended up not containing any slashes but is
+        // actually instead pointing to an existing `null` value then we should use that
+        // `null` value.
+        if (token in this.value && this.value[token] === null) {
+          this.value = null;
+          continue;
+        }
+
         this.value = null;
 
         const path = this.$ref.path || "";
--- a/lib/ref.ts
+++ b/lib/ref.ts
@@ -1,4 +1,4 @@
-import Pointer from "./pointer.js";
+import Pointer, { nullSymbol } from "./pointer.js";
 import type { JSONParserError, MissingPointerError, ParserError, ResolverError } from "./util/errors.js";
 import { InvalidPointerError, isHandledError, normalizeError } from "./util/errors.js";
 import { safePointerToPath, stripHash, getHash } from "./util/url.js";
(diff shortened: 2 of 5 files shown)


jonluca:
This is a great point. 

I think we use `this.value = null` as an explicit check, and don't support the value actually being null. 

Maybe if we change `null` into a `Symbol('null')` and check for that on lines 102 and 124, that's a better solution?

jonluca:
Actually wait, is that a valid schema? It might not be a valid schem. In openapi 3.x, to make a field nullable you set the schema to be `nullable: true`. Does it resolve correctly if you change it to not just be a null value?

erunion:
Yeah that is a valid schema. In this case the thing isn't formally `nullable`, but it's a `null` example value. If I were designing the actual schema for `pas` it'd need to have `nullable` to make this _real_ yeah.

As far as resolving the resolver is able to handle this case without my fix if `pas` is `null` as it just leaves it alone, and changing `pas` to "dog" it also splices it into the schema as a normal `$ref` as well.

Screenshot 2025-02-27 at 1 35 15 PM

It's just when this value we want to splice in is `null` that it breaks down.

```
MissingPointerError: Missing $ref pointer "#/components/examples/product/value/data/pas". Token "pas" does not exist.
 ❯ Pointer.resolve lib/pointer.ts:140:15
    138|         const parentPath = pathFromRoot?.replace(path, "");
    139| 
    140|         throw new MissingPointerError(token, decodeURI(this.originalPath), targetRef, targetFound, parentPath);
       |               ^
    141|       } else {
    142|         this.value = this.value[token];
 ❯ $Ref.resolve lib/ref.ts:122:22
 ❯ $Refs._resolve lib/refs.ts:156:17
 ❯ dereference$Ref lib/dereference.ts:236:25
 ❯ crawl lib/dereference.ts:112:28
 ❯ crawl lib/dereference.ts:155:30
 ❯ crawl lib/dereference.ts:155:30
 ❯ crawl lib/dereference.ts:155:30
 ❯ crawl lib/dereference.ts:155:30
 ❯ crawl lib/dereference.ts:155:30
```

jonluca:
Interesting, that makes sense. 

If we swap out the null checks with a null symbol just for internal tracking that might work. I like your solution for now, I'm fairly open to merging this in. I'll think about it but it feels low risk

erunion:
Doing it as `Symbol('null')` might be tough because when fed through `JSON.stringify()` it becomes `undefined` and consumers will have to recursively parse their schemas to convert them to a true `null` value.

erunion:
This `Symbol` solution works but definitely feels a little on the messy side:

```diff
diff --git a/lib/pointer.ts b/lib/pointer.ts
index f1660a5..d7fa143 100644
--- a/lib/pointer.ts
+++ b/lib/pointer.ts
@@ -125,7 +125,7 @@ class Pointer<S extends object = JSONSchema, O extends ParserOptions<S> = Parser
         // actually instead pointing to an existing `null` value then we should use that
         // `null` value.
         if (token in this.value && this.value[token] === null) {
-          this.value = null;
+          this.value = Symbol('null');
           continue;
         }
 
diff --git a/lib/ref.ts b/lib/ref.ts
index 0ad40b4..f84f8f1 100644
--- a/lib/ref.ts
+++ b/lib/ref.ts
@@ -119,7 +119,12 @@ class $Ref<S extends object = JSONSchema, O extends ParserOptions<S> = ParserOpt
   resolve(path: string, options?: O, friendlyPath?: string, pathFromRoot?: string) {
     const pointer = new Pointer<S, O>(this, path, friendlyPath);
     try {
-      return pointer.resolve(this.value, options, pathFromRoot);
+      const resolved = pointer.resolve(this.value, options, pathFromRoot);
+      if (typeof resolved.value === 'symbol' && resolved.value.toString() === 'Symbol(null)') {
+        resolved.value = null;
+      }
+
+      return resolved;
     } catch (err: any) {
       if (!options || !options.continueOnError || !isHandledError(err)) {
         throw err;
@@ -148,6 +153,9 @@ class $Ref<S extends object = JSONSchema, O extends ParserOptions<S> = ParserOpt
   set(path: string, value: any) {
     const pointer = new Pointer(this, path);
     this.value = pointer.set(this.value, value);
+    if (typeof this.value === 'symbol' && this.value.toString() === 'Symbol(null)') {
+      this.value = null;
+    }
   }
 
   /**
```

jonluca:
For this one I was thinking of having an explicit symbol like

```ts
export const nullSymbol = Symbol('null')
```
and then checking

```ts
if (resolved.value === nullSymbol) {
    resolved.value = null;
}
```

erunion:
🙃 of course, time for afternoon coffee apparently. I've just pushed that refactor up!

jonluca:
Awesome, this is great. Merged in, thanks for the great PR!

PR #381, opened 2025-04-12
fix(dereference): cache poisoning when dereferencing external schemas

erunion:
Apologies for the fast follow here but I uncovered a bug in https://github.com/APIDevTools/json-schema-ref-parser/pull/380 while incorporating v12 into our OpenAPI parser. When dereferencing schemas have external references (eg. `$ref: definitions/child.yaml`) that `$ref` could get saved into the dereference cache as `child.yaml` resulting in the parent schema ending up with an invalid `$ref` pointer that would be `child.yaml`:

screenshot_2025-04-11_at_5 12 33___pm

This case unfortunately was not caught because the `circular-external` test suite did not have a case for `dereference.circular=ignore`.

I **loathe** the fix I am submitting here but it resolves this, very unique, case and as we both agree in https://github.com/APIDevTools/json-schema-ref-parser/pull/380 that the current dereference caching system needs a completely overhaul.

### DIFF
--- a/lib/dereference.ts
+++ b/lib/dereference.ts
@@ -220,43 +220,45 @@ function dereference$Ref<S extends object = JSONSchema, O extends ParserOptions<
     //
     // If the cached object however is _not_ circular and there are additional keys alongside our
     // `$ref` pointer here we should merge them back in and return that.
-    if (cache.circular) {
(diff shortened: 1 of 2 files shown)


lib/dereference.ts
erunion, line 1: Might want to review this with whitespace changes off.

Issue #370, opened 2025-02-04
Recursive (Circular) model $ref doesn't dereference unless wrapped in anyOf (sibling keys alongside $ref keys)


iggyfisk:
[Pydantic recently released an update](https://github.com/pydantic/pydantic/pull/10029) changing their JSON schema generator. 

Previously the generated schema looked like 

```
{
    "$defs": {
        "RecursiveModel": {
            "additionalProperties": false,
            "properties": {
                "value": {
                    "description": "A string",
                    "type": "string"
                },
                "children": {
                    "description": "Children with strings",
                    "items": {
                        "$ref": "#/$defs/RecursiveModel"
                    },
                    "type": "array"
                }
            },
            "required": [
                "value",
                "children"
            ],
            "title": "RecursiveModel",
            "type": "object"
        }
    },
    "allOf": [
        {
            "$ref": "#/$defs/RecursiveModel"
        }
    ]
}
```

which dereferences properly with json-schema-ref-parser. However, now the generated schema looks like

```
{
    "$defs": {
        "RecursiveModel": {
            "additionalProperties": false,
            "properties": {
                "value": {
                    "description": "A string",
                    "type": "string"
                },
                "children": {
                    "description": "Children with strings",
                    "items": {
                        "$ref": "#/$defs/RecursiveModel"
                    },
                    "type": "array"
                }
            },
            "required": [
                "value",
                "children"
            ],
            "title": "RecursiveModel",
            "type": "object"
        }
    },
    "$ref": "#/$defs/RecursiveModel"
}
```
 
which results in an unresolved `'$ref': '#/%24defs/RecursiveModel/properties/children/items'` after `$RefParser.dereference()`. As far as I can tell the [newer schema should be valid](https://www.jsonschemavalidator.net/s/7nBBwmht).

Here's a full script showing the difference, and unresolved `$ref`
```JavaScript
import $RefParser from "@apidevtools/json-schema-ref-parser";

async function main() {
  console.log("----OLD----");
  const old_dereferenced = await $RefParser.dereference({
    "$defs": {
      "RecursiveModel": {
        "additionalProperties": false,
        "properties": {
          "value": {
            "description": "A string",
            "type": "string",
          },
          "children": {
            "description": "Children with strings",
            "items": {
              "$ref": "#/$defs/RecursiveModel",
            },
            "type": "array",
          },
        },
        "required": ["value", "children"],
        "title": "RecursiveModel",
        "type": "object",
      },
    },
[...]

Issue #356, opened 2024-08-14
Relative references using `$def` are being URL encoded


drpump:
We are using a `$defs` section in our schemas. Schemas within the `$defs` section are referenced via local file references, and these schemas can themselves have cross references to other files. The ref parser is correctly injecting a`$defs` cross-reference but replacing the `$defs` with `%24defs` (i.e. it is applying URL encoding). I don’t think this behaviour is correct: while `$` is a reserved character in the URIs RFC, it is designated as a sub-delimiter

For example in the parent file:
```
{
  “properties”: {
     “foo”: “string”
  },
  ”$defs”: {
    “myschema”: {
       “$ref”: “./myschema.json”
    },
    ”otherschema”: {
       “$ref”: “./otherschema.json”
    }
  }
}
```

And the `myschema.json` file:
```
{
      “allOf”: [
        { “$ref”: “./otherschema.json” },
        { 
           “properties”: {
              “someProp”: “string”
           }
        }
      ]
}
```

Flattening results in:

```
{
  “properties”: {
     “foo”: “string”
  },
  ”$defs”: {
    “myschema”: {
      “allOf”: [
        { “$ref”: “#/%24defs/otherschema.” },
        { 
           “properties”: {
              “someProp”: “string”
            }
        }
      ]  
    },
    ”otherschema”: {
         “properties”: {
            “otherProp”: “string”
         }
     }
  }
}
```

lindenquan:
There was the same [issue 262](https://github.com/APIDevTools/json-schema-ref-parser/issues/262).

drpump:
Note that this is occurring in v11.6.4. I haven’t updated to 11.7 yet.

simontaurus:
It also seems like the `$defs` keyword of a schema is not handled like the previous keyword `definitions`, but since this is not enforced and the convention changed over time makes it really tricky.

Code dealing with "definitions"
https://github.com/search?q=repo%3AAPIDevTools%2Fjson-schema-ref-parser+definitions&type=code

Especially:
https://github.com/APIDevTools/json-schema-ref-parser/blob/27abe40702c18c02cae2ad9a0fdd5a33199d9154/lib/bundle.ts#L77

No matching code for "defs"
https://github.com/search?q=repo%3AAPIDevTools%2Fjson-schema-ref-parser+defs&type=code

However, not really related to the encoding of `$` in json-pointers.

KurtGokhan:
[...]

PR #383, opened 2025-04-27
fix: allow $defs and fix circular pointers issue with $ref in root

### LINKED ISSUE #382: Resolving top-level `$ref` fails with `TypeError: Converting circular structure to JSON`
When using `$ref` on the top-level to reference a defined type in `$defs` it wrongly results in a `TypeError: Converting circular structure to JSON` which is not a circular reference.
According to the JSON Schema Core Specification Draft 2020-12 (see `7.5. Applicators`) it is valid to use `$ref` on the top-level.
Here's a example schema:
```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$ref": "#/$defs/TopLevelType",
  "$defs": {
    "PropType": {
      "title": "Property",
      "type": "string"
    },
    "TopLevelType": {
      "properties": {
        "prop1": {
          "$ref": "#/$defs/PropType"
        }
      },
      "title": "Top Level Type",
      "type": "object"
    }
  }
}
```

Trying to resolve this issue using the following test program:
```javascript
import $RefParser from "@apidevtools/json-schema-ref-parser";
import mySchema from "./bundler-demo-non-working.json" with { type: "json" };

try {
  await $RefParser.dereference(mySchema);
  const clonedSchema = await $RefParser.dereference(mySchema, { mutateInputSchema: false });
  console.log(clonedSchema);
} catch (err) {
  console.error(err);
}
```

Results in the following error:
```bash
node test.js 
TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    |     property 'TopLevelType' -> object with constructor 'Object'
    |     property 'properties' -> object with constructor 'Object'
    |     property 'prop1' -> object with constructor 'Object'
    --- property '$defs' closes the circle
    at JSON.stringify (<anonymous>)
    at normalizeArgs (node_modules/@apidevtools/json-schema-ref-parser/dist/lib/normalize-args.js:46:34)
    at $RefParser.dereference (node_modules/@apidevtools/json-schema-ref-parser/dist/lib/index.js:193:54)
    at Object.dereference (node_modules/@apidevtools/json-schema-ref-parser/dist/lib/index.js:190:37)
    at file://test.js:6:41
```

It has been reproduced with version `12.0.1` of `@apidevtools/json-schema-ref-parser`.

The resolution of this issue is required for https://github.com/skriptfabrik/json-schema-bundler/issues/28.

KurtGokhan:
Fixes #382, #370, #356

I think all of the issues above are caused by the same root cause, that `$ref` in root object isn't correctly resolved.

Changes in this PR:
- Removed the short-circuit in `Pointer.resolve`. As I see, it was there to prevent a circular reference. But it causes incorrect behavior. The root of the problem that should be fixed is that Pointer doesn't have a mechanism to detect circular (self-referencing) refs.
  - There is no test case for failing circular refs. But it's a configuration error and cannot be dereferenced anyway. I think it's an advanced use case that doesn't need to be handled.
  - The change in Pointer caused one test to break, which is `circular-external-direct`. At first it seems like a legit test-case but it's actually not. Changed the test case to be more meaningful.
- Added handling for `$defs` in addition to `definitions`
- Used Vitest snapshots for tests. Let me know if there is a reason to keep using the utils other tests are using.
- Ran prettier format


shahmir-oscilar:
Hey-hey! Thanks for this. I was facing the same issue in my code and I ended up using https://www.npmjs.com/package/patch-package to incorporate your changes and they worked out great!

jonluca:
I'm not sure why some of the tests were removed/changed or why the null dereferencing logic was removed. Can you explain that?

jonluca:
Can you expand on the test case not being valid? 

If it's actually invalid we should remove the test case altogether

KurtGokhan:
The test case was like this:

```
# root.yaml
$ref: ./child.yaml#/foo
```

```
# child.yaml
foo:
  $ref: ./root.yaml#/foo
```

And after I fixed the original issue, this was throwing an error like "property foo can't be found". The error is right because if you break it down:

```
root == child.foo
child.foo == root.foo

therefore:

root = root.foo
```

But `root.foo` doesn't exist.

This test was passing before because since `root.yaml` has `$ref` in it's root it was running into the incorrect behavior that was caused by the line I deleted. But the test itself was wrong because `root = root.foo` and since `root.foo` doesn't exist it should error. For this test, I added `root.foo` so the test case should be valid now.

#### What I mean when I said "There is no test case for failing circular refs"

This fix surfaced one more issue that isn't handled. Imagine a case like:

```
root == child.foo
child.foo == root

therefore:

root == root
```

In this repo I guess this is not called circular ref. It's another kind of circular ref I call "self-referencing" ref. I am not sure how this should be handled, or even if it could be handled. But the error could be graceful at least. Right now this case causes infinite recursion. So I guess there can be a mechanism to at least prevent that. That's why I was saying the `Pointer` utility doesn't have a way to detect circular refs.

jonluca:
That makes sense. This looks good. I think the null logic that was removed comes from `https://github.com/APIDevTools/json-schema-ref-parser/pull/374` - as long as the null test cases still pass there, though, I think this is ok? @erunion you worked on that feature, do you have thoughts?

erunion:
Sorry for the delay on getting to this @jonluca  and @KurtGokhan but I think the change makes sense!

KurtGokhan:
Thanks all. Feel free to ping me if there are any issues related to this.

PR #387, opened 2025-06-05
Fix missing location error

AlexHolly:


jonluca:
When would window be defined but window.location doesn't exist?

AlexHolly:
@jonluca 

Hi, thanks for the quick response!

I tested the fix, but it's still not working in my case. The check should be:

`&& typeof location !== "undefined"`

Using just location to check for undefined won't work reliably here, and window.location is also not available in my environment.

Error:
```
ReferenceError: location is not defined
    at Object.cwd (project\node_modules\@apidevtools\json-schema-ref-parser\dist\lib\util\url.js:93:42)
```

jonluca:
Can you try v13.0.3?

I changed it to use explicit window.location instead. It's odd that location isn't on your globalThis but window is. Hopefully this fixes it

JonLuca

AlexHolly:
One more error in case ` window.location.href == 'about:blank'`

Adding  ` && window.location.href !== 'about:blank'` solves the issue

```
TypeError: Invalid URL
    at new URL (node:internal/url:818:25)
    at Object.resolve (project\node_modules\@apidevtools\json-schema-ref-parser\dist\lib\util\url.js:78:25)
```

jonluca:
No I don't think that's the responsibility of this library - I don't think it makes sense to add hardcoded url exceptions.

I'd recommend changing the mocked location at which jsdom is running.

Issue #346, opened 2024-04-23
Possibility to add `onBundle` callback?


graynk:
Hi! This is just an issue to get your opinion on adding a new `onBundle` callback as part of the `options` object, similar to `onDereference`. I'm willing to contribute this, if you're open to adding this feature. Here's some context on why we need it:

We use json-schema-ref-parser to parse and bundle OpenAPI and AsyncAPI files, heavily relying on overriding the resolvers to fetch external references from GitHub.

One issue with that is: if external `$ref` contains `$schema` and `$id` - the `bundle()` function keeps them, but the resulting OpenAPI document fails validation, due to unexpected additional properties. Here's a concrete minimal example:

schema.json
```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "$id": "someId",
  "type": "object",
  "properties": {
    "exampleProperty": {
      "description": "Lorem ipsum and so on",
      "type": "boolean"
    }
  }
}

```

api.yaml
```yaml
openapi: "3.0.3"
info:
  title: Example API
  version: "1.0"
servers:
  - url: https://localhost:3000/v1
paths:
  /example:
    get:
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                $ref: 'schema.json'
```

script.ts
```typescript
import fs from 'node:fs/promises';
import $RefParser from '@apidevtools/json-schema-ref-parser';
import swaggerParser from '@apidevtools/swagger-parser';
import yaml from 'yaml';

const body = await fs.readFile('api.yaml', 'utf8');
const content = yaml.parse(body);
const result = await $RefParser.bundle(content, { resolve: { external: true } });
await fs.writeFile('bundledApi.yaml', yaml.stringify(result));
await swaggerParser.validate(result);
```

bundledApi.yaml
```yaml
openapi: 3.0.3
info:
  title: Example API
  version: "1.0"
servers:
  - url: https://localhost:3000/v1
paths:
  /example:
    get:
      responses:
        "200":
          description: OK
          content:
            application/json:
              schema:
                $schema: http://json-schema.org/draft-07/schema#
                $id: someId
                type: object
                properties:
                  exampleProperty:
                    description: Lorem ipsum and so on
                    type: boolean
```

This fails validation with:
```
SyntaxError: Swagger schema validation failed.
  #/paths/~1example/get/responses/200/content/application~1json/schema must NOT have additional properties
[...]

PR #391, opened 2025-07-02
feat: add `onBundle` callback

aeworxet:
This PR adds functionality, as well as an options object, for the `bundle` method, which consists of:

- The `onBundle` callback.
- The `excludedPathMatcher` callback.

Closes https://github.com/APIDevTools/json-schema-ref-parser/issues/346

### DIFF
--- a/test/specs/bundle-callback/bundle-callback.spec.ts
+++ b/test/specs/bundle-callback/bundle-callback.spec.ts
@@ -13,7 +13,7 @@ describe("Schema with a $ref", () => {
     const options = {
       bundle: {
         onBundle(path, value, parent, parentPropName) {
-          calls.push({ path, value, parent, parentPropName });
+          calls.push(JSON.parse(JSON.stringify({ path, value, parent, parentPropName })));
         },
       },
     } as Options;


jonluca:
Hello,

The logic looks sound, I think you just need to fix the test comparison. The bundle traversal is stable, right? The tests aren't failing based on platform, so just make sure yarn test returns 100% on your local machine.

test/specs/bundle-callback/bundle-callback.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+          },
+          b: {
+            $ref: "#/definitions/b",
+          },
+        },
+        parentPropName: "a",
+      },
+      {
+        path: "#/definitions/a",
+        value: { $ref: "#/definitions/apath: "#/definitions/b",
```
jonluca, line 38: this is missing a comma and closing quote

jonluca:
That is not the issue, the issue is that there is a syntax error in your code. https://github.com/APIDevTools/json-schema-ref-parser/pull/391/files#r2181747158

The original error is that your deep expect is incorrect. https://github.com/APIDevTools/json-schema-ref-parser/actions/runs/16018336530/job/45189942210

If you fix the hardcoded object it should pass.

Please run the tests locally before submitting an update and verify they all succeed.

aeworxet:
Apologies, I didn't push the test I was debugging.
With [this version](https://github.com/APIDevTools/json-schema-ref-parser/pull/391/commits/6e99154245f183d8e25bd7b2053f8c6f39a001ea) of the test, I get correct results with my production app, but a failing test with `yarn test`.

<details>
  <summary>screenshots</summary>

image

image
</details>

aeworxet:
Removed.

jonluca:
Can you change the test to just `calls.push(JSON.parse(JSON.stringify({ path, value, parent, parentPropName })));` instead, to keep it stable? 

Modifying the refs is how the internal bundler works, it's up to the caller of the `onBundle` passer to know that. Or we could change the caller of onBundle to pass a new version of the objects as well.

PR #393, opened 2025-07-15
fix(bundle): don't look for a `$ref` pointer on a null value

erunion:
We had some customers of our CLI tool report issues[^1] this morning of sudden crashes when attempting to validate an OpenAPI definition after we published a new release of our OpenAPI parser[^1] that included an upgrade from v13.0.5 to v14.1.0 with the following error:

```
Cannot read properties of null (reading '$ref')
```

We were able to narrow it down to a regression introduced in https://github.com/APIDevTools/json-schema-ref-parser/pull/391 where it looks for a `$ref` pointer to supply to the `onBundle` callback. The problem our customers had, and we were able to reproduce, encountered with this was on nullish schema examples and this work attempting to look for a `$ref` pointer in it:

Screenshot 2025-07-15 at 3 31 35 PM

The fix I've submitted here is to just wrap this check in a type assertion to ensure that we're only looking for a `$ref` pointer in an object.

[^1]: https://github.com/readmeio/rdme/issues/1306
[^2]: https://npm.im/@readme/openapi-parser

### DIFF
--- a/lib/bundle.ts
+++ b/lib/bundle.ts
@@ -101,8 +101,12 @@ function crawl<S extends object = JSONSchema, O extends ParserOptions<S> = Parse
           crawl(obj, key, keyPath, keyPathFromRoot, indirections, inventory, $refs, options);
         }
 
-        if (value["$ref"]) {
-          bundleOptions?.onBundle?.(value["$ref"], obj[key], obj as any, key);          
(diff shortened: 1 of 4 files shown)


jonluca:
LGTM - thanks!

PR #430, opened 2026-08-13
fix: apply excludedPathMatcher during external resolution

HugoHSun:
Closes #429

## Summary

External references are resolved before the existing bundle and dereference exclusion matchers run. As a result, a literal `$ref` below an excluded path can still be downloaded or throw a resolution error.

This adds `resolve.excludedPathMatcher` so callers can stop traversal before external references are fetched.

The matcher now also receives the value at the current path:

```ts
excludedPathMatcher?(path: string, value?: unknown): boolean;
```

The value is useful when a document contains both intentional internal references and literal external `$ref` properties in the same section. Callers can preserve the literal value without excluding every reference below that path.

Bundling now checks direct child values as well as containers. This ensures a matching `$ref` is skipped before it enters the bundle inventory.

Existing path-only matchers remain compatible because the second argument is optional.

## Testing

- Added external file and HTTP refs below excluded paths to confirm they are not resolved.
- Added bundle coverage for excluded direct child refs.
- Added a value-aware matcher case that resolves an internal ref while preserving an external `$ref` and its sibling properties.


HugoHSun:
@jonluca Thanks for the detailed review. I have addressed each point:

- Resolution now carries a separate root-relative path for matcher calls while retaining the absolute path for URI resolution. Tests cover exact matching, schemas under an `example/` directory, and exclusions inside external documents.
- `DeepPartial` now preserves callable types, with consumer-facing type coverage for contextual matcher parameters, invalid values, and resolver narrowing.
- The options documentation now covers the matcher, value argument, path format, and exclusion behavior.

For the additional edge case, I chose operation-wide, value-based semantics. Matching a path marks that value and its descendants as literal data for the rest of the operation. It does not only skip processing at that one occurrence.

An internal alias can still resolve to the excluded value or access a literal property within it, but any `$ref` inside that value remains literal and is not followed. This stays consistent with the resolution stage, where those references were intentionally not loaded.

I added coverage for internal aliases, excluded descendants, valid and missing literal properties, parser reuse, and circular input values.

